PR #31, opened 2020-02-04
Machine Identity and ACR access

dkistner:
Allow for Azure Shoots to attach to all cluster machines an Azure user-assigned managed identity which could be used to allow the machine or processes running on it to authenticate to other Azure services.

In addition add an option to configure the kubelet(s) of Azure cluster to access private Azure container registries.

:bangbang: This PR require an mcm release containing this: https://github.com/gardener/machine-controller-manager/pull/395 
This will be v0.26.0. Wait until https://github.com/gardener/gardener-extension-provider-azure/pull/30 is merged.

Azure managed identities can be used to give a virtual machine the ability to authenticate to Azure services like ACR, Vault etc. Identities do have by default no access to any Azure service and need to get roles assigned via Azure RBAC to grant them access to the respective services.

There are two types of identities: `system assigned` ones are automatically created and bound to exactly one machine. The `user assigned` ones are created by user separately and can be attached to multiple machines. This PR add **only** support for `user assigned` identities and those need to be created externally.

The user can specify in the Shoot resource via `.spec.provider.infrastructureConfig.identity` the `name` and the `resourceGroupName` of the user assigned managed identity which should be attached to the cluster machines. Exchanging the identity will require a rolling update of the cluster machines.
Furthermore the user can set the `acrAccess` field to true. In this case the azure provider extension will generate an acr configmap. The webhooks for the `original` OperatingSystemConfig will check if the configmap exists and will modify the resource.

**Release note**:

```improvement user
Azure Shoot cluster can now be configured with an user assigned managed identity which can be used to authenticate to other Azure service. In addition Azure Shoots can be configured to pull images from Azure Container Registries by using a passed managed identity. Adding or exchanging the identity will require a rolling update of all worker machines in the Shoot cluster. Please check the documentation https://github.com/gardener/gardener-extension-provider-azure/blob/master/docs/usage-as-end-user.md#infrastructureconfig
```


dkistner:
@rfranzke I have incorporated the most feedback of https://github.com/gardener/gardener-extensions/pull/586. Could you please have another look?

charts/internal/azure-infra/templates/main.tf
```diff
@@ -73,6 +73,13 @@ resource "azurerm_network_security_group" "workers" {
   {{- end}}
 }
 
+{{ if .Values.identity.enabled -}}
```
rfranzke: Nit: You could just use `if .Values.identity` to check this condition. Then you would "save" the `enabled` flag (it's not needed to have it explicitly as far as I can see). Same below.

charts/internal/machineclass/values.yaml
```diff
@@ -7,6 +7,7 @@ machineClasses:
   vnetName: my-vnet
   subnetName: my-subnet-in-my-vnet
   zone: 1
+  identityID: /subscriptions/subscription-id/resourceGroups/resource-group-name/providers/Microsoft.ManagedIdentity/userAssignedIdentities/identity-name
```
rfranzke:
```suggestion
# identityID: /subscriptions/subscription-id/resourceGroups/resource-group-name/providers/Microsoft.ManagedIdentity/userAssignedIdentities/identity-name
```

As the value is optional in the chart I suggest to make it optional here as well.

docs/usage-as-end-user.md
```diff
@@ -64,6 +68,8 @@ Via the `.zoned` boolean you can tell whether you want to use Azure availability
 If you don't use zones then an availability set will be created and only basic load balancers will be used.
 Zoned clusters use standard load balancers.
 
+In the `identity` section you can specify an [Azure user-assigned managed identity](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview#how-does-the-managed-identities-for-azure-resources-work) which should be attached to all cluster worker machines. With `identity.name` you can specify the name of the identity and with `identity.resourceGroup` you can specify the resource group which contain the identity resource on Azure. The identity need to be created by the user upfront (manually, other tooling, ...). Gardener/Azure Extension will only use the referenced one and won't create an identity. Furthermore the identity have to be in the same subscription as the Shoot cluster. Via the `identity.acrAccess` you can configure the worker machines to use the passed identity for pulling from an [Azure Container Registry (ACR)](https://docs.microsoft.com/en-us/azure/container-registry/container-registry-intro).
```
rfranzke:
```suggestion
In the `identity` section you can specify an [Azure user-assigned managed identity](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview#how-does-the-managed-identities-for-azure-resources-work) which should be attached to all cluster worker machines. With `identity.name` you can specify the name of the identity and with `identity.resourceGroup` you can specify the resource group which contains the identity resource on Azure. The identity needs to be created by the user upfront (manually, other tooling, ...). Gardener/Azure Extension will only use the referenced one and won't create an identity. Furthermore, the identity has to be in the same subscription as the Shoot cluster. Via the `identity.acrAccess` you can configure the worker machines to use the passed identity for pulling from an [Azure Container Registry (ACR)](https://docs.microsoft.com/en-us/azure/container-registry/container-registry-intro).
```

pkg/apis/azure/validation/infrastructure.go
```diff
@@ -86,6 +86,10 @@ func ValidateInfrastructureConfig(infra *apisazure.InfrastructureConfig, resourc
 		}
 	}
 
+	if infra.Identity != nil && (infra.Identity.Name == "" || infra.Identity.ResourceGroup == "") {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("identity"), infra.Identity, "specifying an identity require the name of the identity and the resource group which host the identity"))
```
rfranzke:
```suggestion
		allErrs = append(allErrs, field.Invalid(field.NewPath("identity"), infra.Identity, "specifying an identity requires the name of the identity and the resource group which hosts the identity"))
```

docs/usage-as-end-user.md
rfranzke: Also, I would explain that adding/changing/removing the identity results in a rolling update of the nodes (also in the release notes if not yet done).

pkg/internal/infrastructure/terraform.go
```diff
@@ -82,6 +86,9 @@ func ComputeTerraformerChartValues(infra *extensionsv1alpha1.Infrastructure, cli
 			"routeTableName":    TerraformerOutputKeyRouteTableName,
 			"securityGroupName": TerraformerOutputKeySecurityGroupName,
 		}
+		identityConfig = map[string]interface{}{
+			"enabled": false,
```
rfranzke: As mentioned above this might not be needed at all.

pkg/webhook/controlplane/ensurer.go
```diff
@@ -180,16 +185,28 @@ func (e *ensurer) ensureChecksumAnnotations(ctx context.Context, template *corev
...
 	}
 	return opts, nil
 }
 
-func ensureKubeletCommandLineArgs(command []string) []string {
+func (e *ensurer) ensureKubeletCommandLineArgs(ctx context.Context, ectx genericmutator.EnsurerContext, command []string) ([]string, error) {
 	command = extensionswebhook.EnsureStringWithPrefix(command, "--cloud-provider=", "azure")
 	command = extensionswebhook.EnsureStringWithPrefix(command, "--cloud-config=", "/var/lib/kubelet/cloudprovider.conf")
-	return command
+	command = extensionswebhook.EnsureStringWithPrefix(command, "--cloud-config=", "/var/lib/kubelet/cloudprovider.conf")
```
rfranzke: I guess this line was accidentally copied and should not be present here (compare lines 199/200).

pkg/webhook/controlplane/ensurer.go
```diff
@@ -229,3 +246,66 @@ func (e *ensurer) EnsureKubeletCloudProviderConfig(ctx context.Context, ectx gen
...
+// EnsureAdditionalFile ensures additional systemd files
+func (e *ensurer) EnsureAdditionalFiles(ctx context.Context, ectx genericmutator.EnsurerContext, files *[]extensionsv1alpha1.File) error {
+	return e.ensureAcrConfigFile(ctx, ectx, files)
+}
+
+func (e *ensurer) ensureAcrConfigFile(ctx context.Context, ectx genericmutator.EnsurerContext, files *[]extensionsv1alpha1.File) error {
+	// Check if the ACR configmap exists, if not nothing to do.
+	cm, err := e.getAcrConfigMap(ctx, ectx)
+	if err != nil {
+		return nil
```
rfranzke:
```suggestion
		return err
```

rfranzke:
Looks good now, can you please rebase/make the CI happy?

Issue #1, opened 2020-01-23
Enable UDP for egress traffic on Azure


zanetworker:
Currently we have a problem that if  **standard** loadbalancers are used, no frontend ip configuration will be created for UDP by default which in turns will result in blockage for all UDP egress requests (e.g., NTP, DNS,  etc.). 

To fix this we need to support UDP egress traffic. This can be done either via a real NAT GW (not using the LB as a GW), or by modifying any of the shoot services created on the `kube-system` namespace to also have a UDP port. 

cc @dkistner

tim-ebert:
I added a temporary fix for this issue in #32 

This will add a Service called `allow-udp-egress` with a single UPD port to the `kube-system` namespace of Shoots using Azure Standard LoadBalancers (e.g. Zoned Shoot clusters).
It is needed because the Standard LoadBalancers will block all outgoing UDP traffic if there is not at least one open UDP Port.

This workaround will be removed in a future version, once we have a proper implemenation for UDP egress via an Azure NAT Gateway in place.

dkistner:
I would consider the solution with the `allow-udp-egress` as only way to guarantee udp traffic  in all scenarios. We do not enforce the usage of a natgateway and we will get rid of the basic loadbalancer when we replacing availablitysets with vmo. I guess we will then always have a standard loadbalancer, but a natgateway is only there when the user  configure it (e.g egress needs stable ip). Therefore I would like to stay with the udp service solution for all scenarios to ensure that udp egress is working.

/close

PR #32, opened 2020-02-04
Temporarily fix UDP egress for zoned clusters

tim-ebert:
With this PR the azure extensions add a dummy service of type `LoadBalancer` with a UDP port to the Shoot's `kube-system` namespace (zoned Shoots only).
This allows UDP egress from the shoot workers.

This is only a temporary workaround and should be removed once we have a proper implementation for UDP egress via Azure NAT gateways.

ref #1 

cc @rfranzke @dkistner 

**Release note**:

```improvement operator
UDP egress is now enabled for zoned Azure clusters via a temporary workaround. For this the Azure Provider extension installs an additional Service of Type `LoadBalancer` called `udp-egress-dummy` with a single UDP port into the `kube-system` Namespace of the Shoot.
```


charts/internal/controlplane-shoot/charts/udp-egress-dummy/templates/udp-egress-dummy.yaml
```diff
@@ -0,0 +1,21 @@
...
+# this dummy service creates a LoadBalancer with a UDP port to enable outgoing UPD traffic for Shoots using
+# Azure Standard LoadBalancers (see https://github.com/gardener/gardener-extension-provider-azure/issues/1)
+# TODO (dkistner): this workaround should be removed once we have a proper implementation for UDP egress via NAT GW
+apiVersion: v1
+kind: Service
+metadata:
+  name: udp-egress-dummy
+  namespace: kube-system
+  labels:
+    app: dummy
```
rfranzke: Can you add an annotation and explain why we create this? People might see this service/try to delete it.

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -116,6 +116,17 @@ var ccmShootChart = &chart.Chart{
 	},
 }
 
+var udpEgressDummyChart = &chart.Chart{
+	Name: "udp-egress-dummy",
+	Path: filepath.Join(azure.InternalChartsPath, "udp-egress-dummy"),
```
rfranzke: This path looks wrong. Why have we moved the things into a `controlplane-shoot` subfolder in the first place?

charts/internal/controlplane-shoot/charts/udp-egress-dummy/templates/udp-egress-dummy.yaml
```diff
@@ -0,0 +1,21 @@
+{{ if .Values.enabled }}
+# this dummy service creates a LoadBalancer with a UDP port to enable outgoing UPD traffic for Shoots using
+# Azure Standard LoadBalancers (see https://github.com/gardener/gardener-extension-provider-azure/issues/1)
+# TODO (dkistner): this workaround should be removed once we have a proper implementation for UDP egress via NAT GW
```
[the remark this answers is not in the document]
dkistner: +1

tim-ebert:
Sure, will do.
I also want to test my changes again before we merge this, but I'm stuck with some errors in my local setup...

tim-ebert:
I have now tested the changes in my local setup.
Creation and cleanup of the LoadBalancer Service and Azure LoadBalancer rules for UDP work fine and is only done for zoned clusters.
I also verified that NTP timesync is working properly with the fix in place.

So this PR should be fine now, PTAL :)

Issue #43, opened 2020-02-21
NAT Gateway Support


dkistner:
**What would you like to be added**:
Azure will offer soon a NAT service (currently in a private preview) and there are some scenarios when users could need a dedicated nat service e.g. whitelisting scenarios which require a stable ip(s) for egress connections initiated within the cluster.

Currently all egress traffic from a Gardener managed Azure cluster is routed via the cluster load balancer.

As the nat gateway will come with additional costs I would recommend to integrate it optionally and make it configurable for the users.

As the nat gateway require always to have at least one public ip assigned. I would propose to make it possible for users to pass their public ip address(es) or public ip address range(s) to the extension via `.spec.providerConfig.networks.natGateway.ipAddresses[] | .ipAddressRanges[]`. Only if both lists are empty the Gardener extension would create one public ip and assign it to the service.

```
spec:
  type: azure
  ...
  providerConfig:
    apiVersion: azure.provider.extensions.gardener.cloud/v1alpha1
    kind: InfrastructureConfig
    networks:
      natGateway:
        enabled: <true|false>
        ipAddresses:
        - name: my-ip-resource-name
          resourceGroup: my-ip-resource-group
        ipAddressRanges:
        - name: my-iprange-resource-name
          resourceGroup: my-iprange-resource-group
    ...
```

Later on when we go on with multiple Availability Set support we will probably need to make the nat gateway required. In this case the `.spec.providerConfig.networks.natGateway.enabled` need to be always `true`.

**Why is this needed**:
Support scenarios which require a dedicated nat service e.g. whitelisting scenarios.

**Status**
- [x] Step 1 – NatGateway integration with one public ip attached by Gardener (no bring your own ip for the NatGateway in the first step due to terraform-providers/terraform-provider-azurerm#6052) and only for zoned cluster → #50
- [x] Step 2 – Enable bring your own ip for zoned clusters (when terraform-providers/terraform-provider-azurerm#6052 is fixed an a new version of the azurerm Terraform provider is released) → #54 
- [x] ~Step 3 – Enable NatGateway for AvailabilitySet based/non zoned clusters, when Standard Load Balancer is integrated for non zoned clusters (larger effort, due to LoadBalancer migration etc.). This will probably will require to deploy the NatGateway mandatory/non optional in addition to the Standard LoadBalancer.~

cc @vlerenc, @AndreasBurger, @MSSedusch

MSSedusch:
[...]

PR #54, opened 2020-03-18
NatGateway integration – step 2

dkistner:
Allow users to attach their own zonal public ip addresses to the NatGateway of a Shoot cluster.
Allow also to decide in which zone the NatGateway should be deployed to. User provided public ips needs to be zonal and in the same zone as the NatGateway.

Attaching Public IP Prefixes to the NatGateway will currently not be supported as there is no association resource to attach IP Prefixes to the NatGateway in the `azurerm` Terraform provider. Attaching IP Prefixes is currently only possible inline on the NatGateway resource, but as we were forced to write complicated migration logic for public ips when we moved to `azurerm` v2 Terraform provider, we don't want to repeat that for IP Prefixes.  

~~:x: Do not merge until https://github.com/terraform-providers/terraform-provider-azurerm/issues/6052 is fixed and a new Terraformer is integrated that contain a version of the `azurerm` provider which incorporate the fix for the mentioned issue.~~

Part 2 of #43

Fixes partly #43

**Release note**:

```improvement user
It is now possible to attach user provided public ips to the NatGateway of an Azure Shoot and to specify to which zone the NatGateway of the Azure Shoot should be deployed to. Only zonal public ips can be attached to a NatGateway and they need to be in the same zone like the NatGateway. You can find more information [here](https://github.com/gardener/gardener-extension-provider-azure/blob/master/docs/usage-as-end-user.md#infrastructureconfig).
```

```doc user
Changing the zone NatGateway via `.spec.provider.infrastructureConfig.networks.natGateway.zone` while using a managed NatGateway public ip will lead to a recreation of the public ip and you will most likely end up with a different public ip. That mean egress connections will use then a different public ip as before. You can find more information [here](https://github.com/gardener/gardener-extension-provider-azure/blob/master/docs/usage-as-end-user.md#infrastructureconfig).
```


pkg/apis/azure/types_infrastructure.go
```diff
@@ -33,6 +33,14 @@ type InfrastructureConfig struct {
 	Zoned bool
 }
 
+// AzResourceReference contains information to identify a generic Azure resource.
+type AzResourceReference struct {
```
rfranzke: Why not

```suggestion
type AzureResourceReference struct {
```

? :)

pkg/apis/azure/v1alpha1/types_infrastructure.go
```diff
@@ -159,6 +167,12 @@ type VNetStatus struct {
 type NatGatewayConfig struct {
 	// Enabled is an indicator if NAT gateway should be deployed.
 	Enabled bool `json:"enabled"`
+	// IPAddresses is a list of ip addresses which should be assigned to the NAT gateway.
+	// +optional
+	IPAddresses []AzResourceReference `json:"ipAddresses"`
```
rfranzke:
```suggestion
	IPAddresses []AzResourceReference `json:"ipAddresses,omitempty"`
```

pkg/apis/azure/v1alpha1/types_infrastructure.go
```diff
@@ -159,6 +167,12 @@ type VNetStatus struct {
...
+	// IPAddressRanges is a list of ip address ranges/prefixes which should be assigned to the NAT gateway.
+	// +optional
+	IPAddressRanges []AzResourceReference `json:"ipAddressRanges"`
```
rfranzke:
```suggestion
	IPAddressRanges []AzResourceReference `json:"ipAddressRanges,omitempty"`
```

pkg/apis/azure/validation/infrastructure.go
```diff
@@ -104,6 +118,17 @@ func ValidateInfrastructureConfig(infra *apisazure.InfrastructureConfig, nodesCI
 	return allErrs
 }
 
+func validateAzResourceReference(ref *apisazure.AzResourceReference, fieldPath *field.Path) field.ErrorList {
```
rfranzke:
```suggestion
func validateAzureResourceReference(ref *apisazure.AzResourceReference, fieldPath *field.Path) field.ErrorList {
```

pkg/apis/azure/validation/infrastructure.go
```diff
@@ -92,6 +92,20 @@ func ValidateInfrastructureConfig(infra *apisazure.InfrastructureConfig, nodesCI
 	if !infra.Zoned && infra.Networks.NatGateway != nil {
 		allErrs = append(allErrs, field.Invalid(fldPath.Child("networks", "natGateway"), infra.Networks.NatGateway, "NatGateway is currently only supported for zoned cluster"))
 	}
+	if infra.Zoned && infra.Networks.NatGateway != nil {
+		if !infra.Networks.NatGateway.Enabled && (len(infra.Networks.NatGateway.IPAddresses) > 0 || len(infra.Networks.NatGateway.IPAddressRanges) > 0) {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("networks", "natGateway"), infra.Networks.NatGateway, "NatGateway is not enabled but ip addresses or ip ranges are specified"))
+		} else {
+			ipAddressesField := fldPath.Child("networks", "natGateway", "ipAddresses")
+			for i, ip := range infra.Networks.NatGateway.IPAddresses {
+				allErrs = append(allErrs, validateAzResourceReference(&ip, ipAddressesField.Index(i))...)
```
rfranzke:
```suggestion
				allErrs = append(allErrs, validateAzResourceReference(ip.DeepCopy(), ipAddressesField.Index(i))...)
```

pkg/apis/azure/validation/infrastructure.go
```diff
@@ -92,6 +92,20 @@ func ValidateInfrastructureConfig(infra *apisazure.InfrastructureConfig, nodesCI
...
+			}
+			ipAddressRangesField := fldPath.Child("networks", "natGateway", "ipAddressRanges")
+			for i, ipRange := range infra.Networks.NatGateway.IPAddressRanges {
+				allErrs = append(allErrs, validateAzResourceReference(&ipRange, ipAddressRangesField.Index(i))...)
```
rfranzke:
```suggestion
				allErrs = append(allErrs, validateAzResourceReference(ipRange.DeepCopy(), ipAddressRangesField.Index(i))...)
```

pkg/apis/azure/validation/infrastructure_test.go
```diff
@@ -289,6 +289,49 @@ var _ = Describe("InfrastructureConfig validation", func() {
 					"Detail": Equal("NatGateway is currently only supported for zoned cluster"),
 				}))
 			})
+
+			It("should return an error using a NatGateway as it not enabled but ip addresses are configured", func() {
```
rfranzke: I'd also add a test for the ip address ranges here.

docs/usage-as-end-user.md
```diff
@@ -66,7 +66,7 @@ You can freely choose this CIDR and it is your responsibility to properly design
...
... of the Shoot cluster. The NatGateway is currently optional and can be enabled/disabled via the field `networks.natGateway.enabled`. If the NatGateway is not deployed then the outgoing traffic initiated within the Shoot cluster will be routed via cluster LoadBalancer (default behaviour, see [here](https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-outbound-connections#scenarios)). Via `networks.natGateway.ipAddresses` and `networks.natGateway.ipAddressRanges` can own public ip addresses and/or public ip ranges/prefixes defined whose should be attached to the NatGateway. For each public ip addresses or public ip range/prefix need a `.name` (name of the Azure resource) and a `.resourceGroup` (resource group which contains the Azure resource) specified. **Restrictions:** The NatGateway is currently only available for zoned clusters (`.zoned=true`, see [#43](https://github.com/gardener/gardener-extension-provider-azure/issues/43) for more details) and it will not be deployed zone-redundant yet. Furthermore, the Azure NatGateway is not yet generally available (GA) from Azure side, hence, you need to register your subscription to participate in the preview for NatGateway.
```
rfranzke:
```suggestion
The `networks.natGateway` section contains configuration for the Azure NatGateway which can be attached to the worker subnet of the Shoot cluster. The NatGateway is currently optional and can be enabled/disabled via the field `networks.natGateway.enabled`. If the NatGateway is not deployed then the outgoing traffic initiated within the Shoot cluster will be routed via cluster LoadBalancer (default behaviour, see [here](https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-outbound-connections#scenarios)). Via `networks.natGateway.ipAddresses` and `networks.natGateway.ipAddressRanges` you can bring own public ip addresses and/or public ip ranges/prefixes which should be attached to the NatGateway. For each public ip addresses or public ip range/prefix you need to specify the `.name` (name of the Azure resource) and the `.resourceGroup` (resource group which contains the Azure resource). **Restrictions:** The NatGateway is currently only available for zoned clusters (`.zoned=true`, see [#43](https://github.com/gardener/gardener-extension-provider-azure/issues/43) for more details) and it will not be deployed zone-redundant yet. Furthermore, the Azure NatGateway is not yet generally available (GA) from Azure side, hence, you need to register your subscription to participate in the preview for NatGateway.
```

If you put every sentence into its own line it'd be easier to submit suggestions ;-)

pkg/internal/infrastructure/terraform.go
```diff
@@ -171,6 +176,38 @@ func ComputeTerraformerChartValues(infra *extensionsv1alpha1.Infrastructure, cli
 	}, nil
 }
 
+func getNatGatewayValues(natConfig *api.NatGatewayConfig) *map[string]interface{} {
```
rfranzke:
```suggestion
func getNatGatewayValues(natConfig *api.NatGatewayConfig) map[string]interface{} {
```

pkg/internal/infrastructure/terraform.go
```diff
@@ -171,6 +176,38 @@ func ComputeTerraformerChartValues(infra *extensionsv1alpha1.Infrastructure, cli
...
+	if ipAddresses := resourceReferencesToValue(&natConfig.IPAddresses); ipAddresses != nil {
+		values["ipAddresses"] = *ipAddresses
+	}
+	if ipAddressRanges := resourceReferencesToValue(&natConfig.IPAddressRanges); ipAddressRanges != nil {
+		values["ipAddressRanges"] = *ipAddressRanges
+	}
+	return &values
+}
+
+func resourceReferencesToValue(references *[]api.AzResourceReference) *[]map[string]interface{} {
```
rfranzke:
```suggestion
func resourceReferencesToValue(references []api.AzResourceReference) []map[string]interface{} {
```

pkg/internal/infrastructure/terraform.go
```diff
@@ -171,6 +176,38 @@ func ComputeTerraformerChartValues(infra *extensionsv1alpha1.Infrastructure, cli
...
+		return nil
+	}
+	var values = map[string]interface{}{}
+	if ipAddresses := resourceReferencesToValue(&natConfig.IPAddresses); ipAddresses != nil {
+		values["ipAddresses"] = *ipAddresses
+	}
+	if ipAddressRanges := resourceReferencesToValue(&natConfig.IPAddressRanges); ipAddressRanges != nil {
+		values["ipAddressRanges"] = *ipAddressRanges
+	}
+	return &values
```
rfranzke:
```suggestion
	return values
```

pkg/internal/infrastructure/terraform.go
```diff
@@ -171,6 +176,38 @@ func ComputeTerraformerChartValues(infra *extensionsv1alpha1.Infrastructure, cli
...
+}
+
+func resourceReferencesToValue(references *[]api.AzResourceReference) *[]map[string]interface{} {
+	var count = len(*references)
```
rfranzke:
```suggestion
	var count = len(references)
```

pkg/internal/infrastructure/terraform.go
```diff
@@ -171,6 +176,38 @@ func ComputeTerraformerChartValues(infra *extensionsv1alpha1.Infrastructure, cli
...
+		return nil
+	}
+	var result = make([]map[string]interface{}, count)
+	for i, ref := range *references {
+		result[i] = map[string]interface{}{
+			"name":          ref.Name,
+			"resourceGroup": ref.ResourceGroup,
+		}
+	}
+	return &result
```
rfranzke:
```suggestion
	return result
```

pkg/internal/infrastructure/terraform.go
```diff
@@ -171,6 +176,38 @@ func ComputeTerraformerChartValues(infra *extensionsv1alpha1.Infrastructure, cli
...
+	return &values
+}
+
+func resourceReferencesToValue(references *[]api.AzResourceReference) *[]map[string]interface{} {
+	var count = len(*references)
+	if count == 0 {
+		return nil
+	}
+	var result = make([]map[string]interface{}, count)
+	for i, ref := range *references {
```
rfranzke:
```suggestion
	for _, ref := range references {
```

pkg/internal/infrastructure/terraform.go
```diff
@@ -171,6 +176,38 @@ func ComputeTerraformerChartValues(infra *extensionsv1alpha1.Infrastructure, cli
...
+		result[i] = map[string]interface{}{
```
rfranzke:
```suggestion
		result = append(result, map[string]interface{}{
```

pkg/internal/infrastructure/terraform.go
```diff
@@ -171,6 +176,38 @@ func ComputeTerraformerChartValues(infra *extensionsv1alpha1.Infrastructure, cli
...
+			"name":          ref.Name,
+			"resourceGroup": ref.ResourceGroup,
+		}
```
rfranzke:
```suggestion
		})
```

rfranzke (dismissed):
Looks good, a few minor suggestions. :)

pkg/apis/azure/types_infrastructure.go
dkistner: golint wrote:
> type name will be used as azure.AzureResourceReference by other packages, and that stutters;
 
:)

But you are right it should be `AzureResourceReference`. I have changed it.

pkg/apis/azure/v1alpha1/types_infrastructure.go
dkistner: done

pkg/apis/azure/v1alpha1/types_infrastructure.go
dkistner: done

pkg/apis/azure/validation/infrastructure.go
dkistner: done

pkg/apis/azure/validation/infrastructure.go
dkistner: Why do we need a copy of the ip here instead of passing the reference?

pkg/internal/infrastructure/terraform.go
dkistner: done

pkg/apis/azure/types_infrastructure.go
rfranzke: Then call it only `ResourceReference`? :)

pkg/apis/azure/validation/infrastructure.go
rfranzke: `ip` is the iteration variable for the `range`, and it always has the same reference

dkistner: ahh damn always fell on it... Thanks.

dkistner: done

pkg/apis/azure/validation/infrastructure.go
dkistner: done

pkg/apis/azure/validation/infrastructure_test.go
dkistner: yep done

docs/usage-as-end-user.md
dkistner: 😄  yep I will do that.
done.

pkg/internal/infrastructure/terraform.go
dkistner: done

pkg/internal/infrastructure/terraform.go
dkistner: If we do that we would need to init the slice with length 0 as append expand the slice. In combination with `make([]map[string]interface{}, count)` we would get a length of `2x count` and a slice with empty values.

So with append we can use `var result = []map[string]interface{}{}` to init the slice.

But I prefer if the size is already known to init the slice with required size and require the amount of memory directly. Then no copying during appends would be required.

pkg/internal/infrastructure/terraform.go
dkistner: See comment below

pkg/internal/infrastructure/terraform.go
dkistner: done

pkg/internal/infrastructure/terraform.go
dkistner: done

pkg/internal/infrastructure/terraform.go
dkistner: done

rfranzke:
Converted to a draft PR as its dependencies are not yet resolved. Please mark as 'ready for review' once they are resolved.

rfranzke:
https://github.com/terraform-providers/terraform-provider-azurerm/issues/6052 seems now fixed as of https://github.com/terraform-providers/terraform-provider-azurerm/pull/6450.

dkistner:
Nice! Let's wait until there is a new release of the `azurerm` terraform provider. I will try it out then and update the pr accordingly.

dkistner:
Now there is an release which contain the fix https://github.com/terraform-providers/terraform-provider-azurerm/blob/v2.12.0/CHANGELOG.md#2120-may-28-2020
I will try out.

dkistner:
So have tried it out. It does not work without introducing a new association resource for the public ip to nat gateway instead of using the internal public ip association of the nat gateway resource.

The migration from the internal association to association resource does not work, see here: https://github.com/terraform-providers/terraform-provider-azurerm/issues/7167
Seems it can be solved by importing the public ip resource to the Terraform state.

ialidzhikov:
> The migration from the internal association to association resource does not work, see here: [terraform-providers/terraform-provider-azurerm#7167](https://github.com/terraform-providers/terraform-provider-azurerm/issues/7167)
> Seems it can be solved by importing the public ip resource to the Terraform state.

I checked https://github.com/terraform-providers/terraform-provider-azurerm/issues/7167 and it is quite unfortunate.
The thing I am wondering is whether we can have have

```tf

resource "azurerm_nat_gateway" "nat" {
  name                    = "{{ required "clusterName is required" .Values.clusterName }}-nat-gateway"
  location                = "{{ required "azure.region is required" .Values.azure.region }}"
  {{ if .Values.create.resourceGroup -}}
  resource_group_name     = azurerm_resource_group.rg.name
  {{- else -}}
  resource_group_name     = data.azurerm_resource_group.rg.name
  {{- end }}
  sku_name                = "Standard"
  {{ if .Values.natGateway.migrateToPublicIPAssociation -}}
  public_ip_address_ids   = []
  {{- end }}
}

resource "azurerm_nat_gateway_public_ip_association" "nat-ip-association" {
   nat_gateway_id       = "${azurerm_nat_gateway.nat.id}"
   public_ip_address_id = "${azurerm_public_ip.natip.id}"
}

```

I guess we could have in the Infrastructure status field like `migratedToPublicIPAssociation` that indicates whether the infra migrated to `azurerm_nat_gateway_public_ip_association`.
In the first apply run, the terraformer would run with `migrateToPublicIPAssociation: true` which will remove for a short the public IP addresses from the nat gateway (because  of `public_ip_address_ids   = []`) but then it will add them once again because of the `azurerm_nat_gateway_public_ip_association` resource. After successful run, we could set the `status.migratedToPublicIPAssociation=true` and then pass in the subsequent runs `migrateToPublicIPAssociation=false` to the terraformer chart.
I don't know exactly what is the complexity of the `terraform import` way but this approach sounds easier to implement? WDYT?

dkistner:
@ialidzhikov I had the same idea and already implemented that almost exactly like you proposed. The issue here is that there will be a time during the first and the second reconciliation where no public ip is assigned to the NatGateway. In this period of time there is no ip assigned to the NatGateway and it can't be used to route egress traffic which require a stable ip. For some workloads this would mean a serious disruption.

Can we somehow influence that an infrastructure resource get reconciled multiple times even after the previous reconcile succeed? I didn't find a way to archive that. Therefore I rather thought in the direction of import support.

I think having a generic way to support Terraform imports for the Gardener extensions could also help in the future with other migration scenarios.

ialidzhikov:
> @ialidzhikov I had the same idea and already implemented that almost exactly like you proposed. The issue here is that there will be a time during the first and the second reconciliation where no public ip is assigned to the NatGateway. In this period of time there is no ip assigned to the NatGateway and it can't be used to route egress traffic which require a stable ip. For some workloads this would mean a serious disruption.

Hmm, why there would be 2 reconciliations. Setting `public_ip_address_ids` to `[]` and introduction of `azurerm_nat_gateway_public_ip_association ` happen in the same infra reconciliation - one after other in the terraform execution. So the public ip should be removed and then associated again after few seconds.

This is what we have now:

```tf

resource "azurerm_public_ip" "natip" {
  name                = "nat-ip"
  location            = "westeurope"
  resource_group_name = "${azurerm_resource_group.rg.name}"
  allocation_method   = "Static"
  sku                 = "Standard"
}

resource "azurerm_nat_gateway" "nat" {
  name                    = "nat-gateway"
  location                = "westeurope"
  resource_group_name     = "${azurerm_resource_group.rg.name}"
  sku_name                = "Standard"
  public_ip_address_ids   = ["${azurerm_public_ip.natip.id}"]
}

```

This is how migration runs looks like
```tf

resource "azurerm_public_ip" "natip" {
  name                = "nat-ip"
  location            = "westeurope"
  resource_group_name = "${azurerm_resource_group.rg.name}"
  allocation_method   = "Static"
  sku                 = "Standard"
}

resource "azurerm_nat_gateway" "nat" {
  name                    = "nat-gateway"
  location                = "westeurope"
  resource_group_name     = "${azurerm_resource_group.rg.name}"
  sku_name                = "Standard"
  public_ip_address_ids   = []
}

resource "azurerm_nat_gateway_public_ip_association" "nat-ip-association" {
   nat_gateway_id       = "${azurerm_nat_gateway.nat.id}"
   public_ip_address_id = "${azurerm_public_ip.natip.id}"
}
```

This is how a normal run after the migration looks like:

```tf

resource "azurerm_nat_gateway" "nat" {
  name                    = "nat-gateway"
  location                = "westeurope"
  resource_group_name     = "${azurerm_resource_group.rg.name}"
  sku_name                = "Standard"
}

resource "azurerm_nat_gateway_public_ip_association" "nat-ip-association" {
   nat_gateway_id       = "${azurerm_nat_gateway.nat.id}"
   public_ip_address_id = "${azurerm_public_ip.natip.id}"
}

```

dkistner:
Finally, after almost a year, we can now continue with this PR. 
I have completely overhauled the PR and we can now go on a review.
Also updated the PR description with the actual information.

/touch
/invite @kon-angelo @ialidzhikov

dkistner:
Thanks @ialidzhikov 
I have addressed the requested changes.

ialidzhikov (approved):
/lgtm
/needs second-opinion

pkg/apis/azure/validation/infrastructure.go
```diff
@@ -68,23 +68,72 @@ func ValidateInfrastructureConfig(infra *apisazure.InfrastructureConfig, nodesCI
...
 }
 
+func validateNatGatewayIPReference(references []apisazure.PublicIPReference, zone int32, fldPath *field.Path) field.ErrorList {
+	var allErrs = field.ErrorList{}
+	for i, ref := range references {
+		if ref.Zone != zone {
+			allErrs = append(allErrs, field.Invalid(fldPath.Index(i).Child("zone"), ref.Zone, fmt.Sprintf("Public IP can't be used as it is not in the same zone as the NatGateway (zone %d)", zone)))
+		}
+		if ref.Name == "" {
+			allErrs = append(allErrs, field.Invalid(fldPath.Index(i).Child("name"), ref.Name, "Empty name for NatGateway public IP is invalid"))
```
kon-angelo, line 128: Nitpick, but shouldn't the error here be of type `field.Required` ? You are anyway trying to print an empty string for the name here.

pkg/apis/azure/validation/infrastructure.go
```diff
@@ -68,23 +68,72 @@ func ValidateInfrastructureConfig(infra *apisazure.InfrastructureConfig, nodesCI
...
+		}
+		if ref.ResourceGroup == "" {
+			allErrs = append(allErrs, field.Invalid(fldPath.Index(i).Child("resourceGroup"), ref.ResourceGroup, "Empty resourceGroup for NatGateway public IP is invalid"))
```
kon-angelo, line 131: Same as comment above: maybe change to `field.Required`.

pkg/apis/azure/validation/infrastructure.go
kon-angelo: Two of the tests will also need to be adapted accordingly to change the error assertion.

dkistner: Yes agree they should rather be required then invalid.

dkistner:
@kon-angelo Thanks for the review. I have addressed the requested changes.

kon-angelo (approved):
/lgtm

PR #58, opened 2020-03-22
Add support for VM image ID

vpnachev:
Add support for VM images that do not support URN, for example private images.

Fixes #

The feature depends on https://github.com/gardener/machine-controller-manager/pull/436

**Release note**:

```improvement operator
It is now possible to use private or other VM images which cannot be referred with URN. To use such an image the image id has to be specified in `cloudprofiles.spec.providerConfig.machineImages[].versions[].id` field.
```

```noteworthy operator
The VM image ID feature is supported only with [MCM](https://github.com/gardener/machine-controller-manager) version >=v0.27.x
```


vpnachev:
This PR is now ready for review.

pkg/apis/azure/helper/helper.go
```diff
@@ -101,11 +101,22 @@ func FindImageFromCloudProfile(cloudProfileConfig *api.CloudProfileConfig, image
 			}
 			for _, version := range machineImage.Versions {
 				if imageVersion == version.Version {
-					urn := version.URN
+					if (version.URN == nil && version.ID == nil) || (version.URN != nil && version.ID != nil) {
```
rfranzke: Is this double validation really needed? You already validate the same in the `pkg/apis/azure/validation`.

vpnachev: I was wondering the same and decided to put it in the PR to get some feedback.
So I will remove it.

vpnachev: removed

dkistner:
Hmm we have merged that, but we don't have a new mcm release which include https://github.com/gardener/machine-controller-manager/pull/436. Therefore we are currently not able to release a new version of the extension...

cc @vpnachev, @rfranzke

rfranzke:
I see... we'll revert it. I think it makes sense to only create DRAFT PRs if there are external dependencies... I missed the comment in the PR description. Thanks Dominic.

vpnachev:
It is not necessary to revert this PR.
By default the image ID feature will not be available, but the MCM can be upgraded via imageVectorOverwrite.

In a PM @hardikdr told me he is planning to release MCM this week, so let's give him some time.

rfranzke:
I think it's still safer to revert it unless the MCM version in `charts/images.yaml` has been bumped. Otherwise, the Azure APIs expose configuration options that are not supported, leading to bad user experience. Advising end-users to use the image vector will not improve this experience. :(
A new release is not planned before next week - so we can wait - maybe a new MCM release makes it until then.

prashanth26:
Hi Guys, 

We are planning to make a release today. Positively by 12pm CET. Give us some time until then.

CC: @hardikdr

hardikdr:
Yes, we are releasing MCM today.

rfranzke:
Awesome, thanks for the update colleagues! :) Then everything should be fine.

PR #66, opened 2020-04-01
Revert "Add support for VM image ID"

vpnachev:
Reverts gardener/gardener-extension-provider-azure#58

It turns out the images IDs are different for every region, thus #58 will work only if the shoot clusters are created in the same region where the image is located, which was not what I wanted.

Ref: https://devblogs.microsoft.com/cse/2018/02/15/copy-custom-vm-images-on-azure/


vpnachev:
I'm working on a fix, but if I am not ready before the release of the extension, please merge this PR.

dkistner:
Thanks @vpnachev.

PR #99, opened 2020-05-28
Switch from cm to secret for cloud provider config

timuthy:

/area control-plane
/kind enhancement
/priority normal
/platform azure

This PR let's the extension create/use Secrets instead of ConfigMaps for `cloud-provider-config` and `cloud-provider-disk-config`.

**Release note**:

```improvement operator
The Kubernetes Kind for `cloud-provider-config` and `cloud-provider-disk-config` in a shoot's control plane has been changed from ConfigMap to Secret.
```

### DIFF
--- a/pkg/controller/controlplane/valuesprovider.go
+++ b/pkg/controller/controlplane/valuesprovider.go
@@ -42,6 +42,7 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	storagev1beta1 "k8s.io/api/storage/v1beta1"
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apiserver/pkg/authentication/user"
 	autoscalingv1beta2 "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1beta2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -163,11 +164,11 @@ var (
 		Path: filepath.Join(azure.InternalChartsPath, "cloud-provider-config"),
 		Objects: []*chart.Object{
 			{
-				Type: &corev1.ConfigMap{},
+				Type: &corev1.Secret{},
 				Name: azure.CloudProviderConfigName,
 			},
 			{
-				Type: &corev1.ConfigMap{},
+				Type: &corev1.Secret{},
 				Name: azure.CloudProviderDiskConfigName,
 			},
 		},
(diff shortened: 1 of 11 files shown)


pkg/controller/controlplane/valuesprovider.go
```diff
@@ -341,9 +342,41 @@ func (vp *valuesProvider) GetConfigChartValues(
...
 
+func deleteLegacyCloudProviderConfigMap(ctx context.Context, c client.Client, namespace string) error {
+	ccmDeployment := &appsv1.Deployment{}
+	cm := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      azure.CloudProviderConfigName,
+			Namespace: namespace,
+		},
+	}
+	if err := c.Get(ctx, kutil.Key(namespace, azure.CloudControllerManagerName), ccmDeployment); client.IgnoreNotFound(err) != nil {
```
rfranzke, line 353: Is this really necessary, i.e., can't we simply always try to delete the config map and ignore the 'not found' error, similar to your `deleteLegacyCloudProviderDiskConfigMap` function? We don't really safe anything by first reading the CCM deployment, do we?
OTOH, I don't have any strong opinion, I didn't want to comment at all, but then I saw the `deleteLegacyCloudProviderDiskConfigMap` implementation and was wondering why there isn't similar code that determines whether the config map should be deleted.

pkg/webhook/controlplane/ensurer.go
```diff
@@ -300,6 +300,11 @@ func (e *ensurer) ensureChecksumAnnotations(ctx context.Context, template *corev
 		return nil
 	}
 
+	if err := controlplane.EnsureSecretChecksumAnnotation(ctx, template, e.client, namespace, azure.CloudProviderConfigName); err == nil {
+		return nil
```
rfranzke, line 304: Shouldn't we handle potential errors here?

pkg/webhook/controlplane/ensurer.go
```diff
@@ -392,18 +397,34 @@ func (e *ensurer) ShouldProvisionKubeletCloudProviderConfig(ctx context.Context,
...
+	}
+
+	if len(secret.Data[azure.CloudProviderConfigMapKey]) > 0 {
+		*data = string(secret.Data[azure.CloudProviderConfigMapKey])
+		return nil
+	}
+
+	if !apierrors.IsNotFound(secretErr) && len(secret.Data[azure.CloudProviderConfigMapKey]) == 0 {
+		return nil
+	}
```
rfranzke, line 413: Hm, what is this condition for?

timuthy: It's this [case](https://github.com/gardener/gardener-extension-provider-azure/blob/3b7440e45f4905430348e6798d17f487f2a01a66/pkg/webhook/controlplane/ensurer.go#L405-L408) but for the secret instead of the config map.

pkg/controller/controlplane/valuesprovider.go
timuthy: Okay, I agree that this check doesn't make too much sense.

PR #100, opened 2020-05-28
Diable accelerated network for AvSet clusters

vpnachev:

/area networking
/kind bug
/priority critical
/platform azure

Disable accelerated networking for AvSet clusters. This is necessary because for legacy clusters it is not possible to create new VMs. The creation calls are failing with 
```text
ExistingAvailabilitySetWasNotDeployedOnAcceleratedNetworkingEnabledCluster" Message="Cannot add a virtual machine (/subscriptions/...) with accelerated-networking-enabled network interface(s) on an existing availability set (/subscriptions/...) with VM(s) without accelerated-networking-enabled network interfaces.
``` 

Fixes #

See https://github.com/MicrosoftDocs/azure-docs/issues/10536 for more details.

**Release note**:

```noteworthy user
Accelerated Networking feature has been disabled for AvSet clusters. It is still available for zoned clusters. 
```

### DIFF
--- a/docs/usage-as-end-user.md
+++ b/docs/usage-as-end-user.md
@@ -223,5 +223,6 @@ Shoot clusters with Kubernetes v1.18 or less will use the in-tree `kubernetes.io
 ### Azure Accelerated Networking
 
 All worker machines of the cluster will be automatically configured to use [Azure Accelerated Networking](https://docs.microsoft.com/en-us/azure/virtual-network/create-vm-accelerated-networking-cli) if the prerequisites are fulfilled.
-The prerequisites are that the used machine type and operating system image version are compatible for Accelerated Networking.
+The prerequisites are that the cluster must be zoned, and the used machine type and operating system image version are compatible for Accelerated Networking.
+`Availability Set` based shoot clusters will not be enabled for accelerated networking even if the machine type and operating system support it, this is necessary because all machines from the availability set must be scheduled on special hardware, more daitls can be found [here](https://github.com/MicrosoftDocs/azure-docs/issues/10536).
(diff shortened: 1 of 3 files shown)


docs/usage-as-end-user.md
```diff
@@ -223,5 +223,5 @@ Shoot clusters with Kubernetes v1.18 or less will use the in-tree `kubernetes.io
 ### Azure Accelerated Networking
 
 All worker machines of the cluster will be automatically configured to use [Azure Accelerated Networking](https://docs.microsoft.com/en-us/azure/virtual-network/create-vm-accelerated-networking-cli) if the prerequisites are fulfilled.
-The prerequisites are that the used machine type and operating system image version are compatible for Accelerated Networking.
+The prerequisites are that the cluster must be zoned and the used machine type and operating system image version are compatible for Accelerated Networking. `Availability Set` based shoot clusters will not be enabled for accelerated networking even if the machine type and operating system support it.
```
rfranzke, line 226: Can you add more information here, i.e., the reason why we don't do respect the settings for AvSet clusters?

vpnachev: done

dkistner:
Oh thats a pity. Once more on case which was not part of the test matrix... To many deployment scenarios on Azure...

So we can reenable it when we have an own avset for each workerpool. These avset contain then only machines of same type and configuration (e.g. operating system).

vpnachev:
> So we can reenable it when we have an own avset for each workerpool. These avset contain then only machines of same type and configuration (e.g. operating system).

I am not sure - if the worker pool is rolled from disabled accelerated network to enabled, we will face the same issue. So we have to keep status of the AvSet and disallow accelerated networking if it still has NICs with disabled accelerated networking. Which means we will have to list all NICs in the AvSet which is one more operation contributing to the Azure API calls.

vpnachev:
Valid point, I've opened #101 to summarize the topic and keep track on it there.

PR #111, opened 2020-06-16
Retain fault and update domain counts

### LINKED ISSUE #109: Infra failure when fault or update domain count is changed

/kind bug
/priority normal
/platform azure

**What happened**:
The values `countFaultDomains` or `countUpdateDomains` were changed in a cloud profile. A corresponding shoot had been created successfully with the previous values but when Gardener tried to reconcile the `Infrastructure` of the shoot with the new values, the infra controller failed for the following reason:

```
* compute.AvailabilitySetsClient#Delete: Failure sending request: StatusCode=409 -- Original Error: autorest/azure: Service returned an error. Status=<nil> Code="OperationNotAllowed" Message="Availability Set '<technical-id>' cannot be deleted. Before deleting an Availability Set please ensure that it does not contain any VM."
```

**What you expected to happen**:
The infra should reconcile successfully, probably with the values which had been used to create the Availability Set.

timuthy:

/area quality
/kind bug
/priority normal
/platform azure

The PR retains the fault and update domain counts for the worker availability set instead of always taking the configured value from the cloud profile.

Fixes #109

~This PR needs https://github.com/gardener/gardener/pull/2460 as a prerequisite. I will re-vendor g/g as soon as the PR is merged.~ 
✅ Bumped to g/g v1.6.2

**Release note**:

```improvement operator
An issue has been fixed which caused a reconcile error for the infrastructure if the `countFaultDomains` or `countUpdateDomains` values for the shoot's region was changed.
```


pkg/internal/infrastructure/terraform.go
```diff
@@ -342,3 +353,51 @@ func ComputeStatus(tf terraformer.Terraformer, config *api.InfrastructureConfig)
...
+			return err
+		}
+		nodesAvailabilitySet, err := helper.FindAvailabilitySetByPurpose(infrastructureStatus.AvailabilitySets, api.PurposeNodes)
+		if err != nil {
+			return err
+		}
+
+		if nodesAvailabilitySet != nil {
+			faultDomainCount = nodesAvailabilitySet.CountFaultDomains
+			updateDomainCount = nodesAvailabilitySet.CountUpdateDomains
```
dkistner, line 375: We can already return here, right? We have found the domain counts in the infra status. No need to check also the cloud profile.

```suggestion
			val["countFaultDomains"] = *nodesAvailabilitySet.CountFaultDomains
			val["countUpdateDomains"] = *nodesAvailabilitySet.CountUpdateDomains
			return nil
```

pkg/apis/azure/types_cloudprofile.go
```diff
@@ -40,7 +40,7 @@ type DomainCount struct {
 	// Region is a region.
 	Region string
 	// Count is the count value for the respective domain count.
-	Count int
+	Count int32
```
dkistner, line 43: Why we need to convert the type to `int32`?

timuthy: It's a K8s API convention: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types

pkg/internal/infrastructure/terraform.go
timuthy: Hm, I don't think so. When we introduce this change, we won't have these fields in the status during the first reconciliation.

dkistner: Ah right, yes for the case when we have already an infra status but the domain counts are not part of it.
Thanks for the explanation :)

rfranzke: Nice catch. Probably a comment would help here - I was about to suggest the same like @dkistner.

pkg/internal/infrastructure/terraform.go
```diff
@@ -116,22 +123,9 @@ func ComputeTerraformerChartValues(infra *extensionsv1alpha1.Infrastructure, cli
 		outputKeys["availabilitySetID"] = TerraformerOutputKeyAvailabilitySetID
 		outputKeys["availabilitySetName"] = TerraformerOutputKeyAvailabilitySetName
 
-		cloudProfileConfig, err := helper.CloudProfileConfigFromCluster(cluster)
-		if err != nil {
-			return nil, err
+		if err := fillDomainCounts(azure, cluster, infra); err != nil {
```
rfranzke, line 126: Style-wise, I would not pass the `azure` and let the function rather return the two values and write it into `azure` here. Then there is only "one place"/function where the Helm chart values are computed. Now you have to look into multiple functions to find where a certain value is coming from. No strong opinion/request to change, though.

timuthy: Okay, agree 👍

PR #120, opened 2020-06-24
Add Azure remedy controller to the controlplane extension

### LINKED ISSUE #119: Add Azure remedy controller to the controlplane extension
/area control-plane
/kind enhancement
/priority normal
/platform azure

Extend the controlplane controller to install / remove the Azure remedy controller.

**Why is this needed**:
The Azure remedy controller takes care about certain sporadic Azure issues by trying to apply "remedies" for them.

stoyanr:
/area control-plane
/kind enhancement
/priority normal
/platform azure

Extends the controlplane controller to install / remove the Azure remedy controller. The Azure remedy controller takes care about certain sporadic Azure issues by trying to apply "remedies" for them.

Fixes #119 

**Release note**:
```improvement operator
The Azure controlplane now contains an additional controller for applying "remedies" for certain sporadic Azure issues.
```


pkg/controller/controlplane/valuesprovider.go
```diff
@@ -48,6 +48,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	// enableRemedyControllerAnnotation enables the Azure remedy controller (disabled by default)
+	// This annotation and the corresponding code should be removed once the remedy controller is considered
+	// fully enabled for production use.
+	enableRemedyControllerAnnotation = "extensions.gardener.cloud/enable-remedy-controller-azure"
```
rfranzke, line 55:
```suggestion
	enableRemedyControllerAnnotation = "azure.provider.extensions.gardener.cloud/enable-remedy-controller"
```

charts/internal/seed-controlplane/charts/remedy-controller-azure/templates/deployment.yaml
```diff
@@ -0,0 +1,87 @@
...
+        app: remedy-controller-azure
+        networking.gardener.cloud/to-dns: allowed
+        networking.gardener.cloud/to-public-networks: allowed
+        networking.gardener.cloud/to-private-networks: allowed
+        networking.gardener.cloud/to-seed-apiserver: allowed
+        networking.gardener.cloud/to-shoot-apiserver: allowed
+        networking.gardener.cloud/from-prometheus: allowed
+    spec:
+      serviceAccountName: remedy-controller-azure
+      tolerations:
```
dkistner, line 33: Why do we need this toleration? Is the remedy controller that important that we also schedule it on nodes with a `NoExecute` taint? If yes do we need maybe a `tolerationSeconds` in case the taint after the pod is already running on the pod?

stoyanr: That's a good question, I am not sure myself. I noticed that this toleration is present for the `cloud-controller-manager` for Azure and other platforms, as well as well as on the AWS LB readvertiser, and I thought the remedy controller was of comparable importance, but this was just a guess and so perhaps is wrong. Perhaps the interesting question is, why anything that runs in the seed cluster, including the components above, would need this toleration? I would understand it for system components running in the shoot, e.g. the CSI node driver. I also remember this is explicitly stated in the CCM documentation, if CCM runs on the cluster nodes (CCM itself removes the taint when a node is added, so if CCM doesn't tolerate the taint, it can't run in order to do this), but in the Gardener case, I believe this doesn't apply? If there is a good reason, perhaps it applies to the remedy controller. If not, perhaps we should remove it from other places as well.

dkistner: Ok then we should maybe also remove the toleration from the other components which @stoyanr mentioned. Thanks for clarification.

stoyanr: OK, I will remove it and if you don't mind, will fix Azure CCM as well. Should I open PRs for the other providers with this change?

rfranzke: Thanks, if you have the time, sure. But I think it's not critical, so go ahead as it fits your schedule.

charts/internal/seed-controlplane/charts/remedy-controller-azure/templates/deployment.yaml
```diff
@@ -0,0 +1,84 @@
...
+        checksum/configmap-remedy-controller-azure-config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+{{- if .Values.podAnnotations }}
+{{ toYaml .Values.podAnnotations | indent 8 }}
+{{- end }}
+      labels:
+        garden.sapcloud.io/role: controlplane
+        app: remedy-controller-azure
+        networking.gardener.cloud/to-dns: allowed
+        networking.gardener.cloud/to-public-networks: allowed
+        networking.gardener.cloud/to-private-networks: allowed
```
danielfoehrKn, line 27: I do not know the details of the remedy controller - could you quickly outline why it needs access to private ip ranges?

stoyanr: The remedy controller needs to be able to talk to both API servers (Seed and Shoot), as well as the Azure API. I believe it needs then access to both public and private networks? Or am I wrong?

charts/images.yaml
```diff
@@ -59,3 +59,8 @@ images:
   sourceRepository: github.com/kubernetes-csi/livenessprobe
   repository: quay.io/k8scsi/livenessprobe
   tag: "v2.0.0"
+
+- name: remedy-controller-azure
+  sourceRepository: github.wdf.sap.corp/kubernetes/remedy-controller
```
AndreasBurger, line 64: Needs an update now:
```suggestion
  sourceRepository: github.com/gardener/remedy-controller
```

PR #131, opened 2020-07-09
Be pessimistic about legacy cm deletion

timuthy:

/area control-plane
/kind bug
/priority normal
/platform azure

With the introduction of #99 we observed errors during the shoot deletion because the KAS deployment still referred to the legacy cloud-provider config map. This PR adds a check if the said config map is still referenced and skips the deletion. Please note that the deletion error only happens under special circumstances, e.g. if the shoot reconciliation is aborted after the control plane deployment and immediately deleted.

/cc @danielfoehrKn

**Release note**:

```improvement operator
An issues has been fixed which caused unsuccessful shoot deletions due to the migration of the `cloud-provider-config` from a config map to a secret.
```


dkistner:
@danielfoehrKn Can you have look?

PR #161, opened 2020-09-09
Prepare upgrade to Terraform azurerm v2.x.x

dkistner:
Add association resources for `routetable-to-subnet` and `nsg-to-subnet` in parallel to internal associations on the subnet resource.

/kind impediment
/priority normal
/platform azure

We need to upgrade the `azurerm` Terraform provider to at least version v2.12.0 as this is required for the NatGateway bring your own IP scenario (https://github.com/gardener/gardener-extension-provider-azure/pull/54#issuecomment-635788184, https://github.com/terraform-providers/terraform-provider-azurerm/pull/6450, https://github.com/terraform-providers/terraform-provider-azurerm/issues/6052). 

As preparation for the upgrade we need to add the `routetable-to-subnet` and `nsg-to-subnet` external association resources in parallel to the internal associations on the subnet resource. Otherwise we won't be able to upgrade to `azurerm` v2, see here: 
 https://github.com/terraform-providers/terraform-provider-azurerm/issues/2358#issuecomment-440168871

Create a `Infrastructure` without this change. Then checkout this PR and reconcile the `Infrastructure` once more. The route table/nsg to subnet association on Azure should not change. 

**Release note**:

```NONE

```

### DIFF
--- a/charts/internal/azure-infra/templates/main.tf
+++ b/charts/internal/azure-infra/templates/main.tf
@@ -63,6 +63,11 @@ resource "azurerm_route_table" "workers" {
   {{- end}}
 }
 
+resource "azurerm_subnet_route_table_association" "workers-rt-subnet-association" {
+  subnet_id      = azurerm_subnet.workers.id
+  route_table_id = azurerm_route_table.workers.id
+}
+
 resource "azurerm_network_security_group" "workers" {
   name                = "{{ required "clusterName is required" .Values.clusterName }}-workers"
   location            = "{{ required "azure.region is required" .Values.azure.region }}"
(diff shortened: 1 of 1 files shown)


ialidzhikov:
Is this PR dependent to the azurerm version bump in terraformer? In the past we had issue with https://github.com/gardener-attic/gardener-extensions/pull/331 which was reverted in https://github.com/gardener-attic/gardener-extensions/pull/348. Is this now fixed?

dkistner:
It is only fixed if you keep the inline and external association in parallel.
If you only use the external one it will alternating attach and detach the routetable and nsg,
With both this should not happen.

This PR is not related to a bump of Terraform version. It is just a preparation that we can move to azurerm v2 in the future (which we need to proceed with the Nat Gateway story).

ialidzhikov (approved):
/lgtm

PR #181, opened 2020-10-22
add priority class for extension pods

danielfoehrKn:

/area high-availability
/kind enhancement
/priority normal
/platform azure

Adds priority class for extension pods to prevent preemption.
Has same priority as the gardenlet class.

Fixes #

We do not know which version the Seed is running on

**Release note**:

```improvement operator
Adds priority class for extension pods to prevent preemption.
```


rfranzke:
Please run `make generate`

rfranzke (approved):
/lgtm

timebertt (approved):

charts/gardener-extension-provider-azure/templates/priorityclass.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: {{ include "priorityclassversion" . }}
+kind: PriorityClass
+metadata:
+  name: gardener-extension-provider-azure
+value: 1000000000
```
timebertt, line 5: Hmm, this is the same value, that the gardenlet has. Should we give the extensions less priority than the gardenlet?

rfranzke: I think it makes sense to use the same value. The extension are similarly important like the gardenlet.

dkistner (approved):

charts/gardener-extension-provider-azure/templates/priorityclass.yaml
timebertt: Ok, good :)

PR #189, opened 2020-11-12
Changing the default parameters handling and adding all properties to the test definition 

kris94:
How to categorize this PR?
/kind enhancement
/priority normal
/platform azure

What this PR does / why we need it:
This PR adds all properties to the "provider-azure" test definition. In addition to this, as discussed in [PR:191,](https://github.com/gardener/gardener-extension-provider-aws/pull/191) the default parameters handling is changed.
This way one test definition will be used by the integration tests where shoots with different networks from the defaults must be created (e.g. for registering a new seed for the control plane migration) and it will preserve the default behaviour for other integration tests (e.g. for shoot creation) that relies on the default values.
Which issue(s) this PR fixes:
Fixes #

Special notes for your reviewer:

Release note:
```
NONE
```

### DIFF
--- a/test/tm/generator.go
+++ b/test/tm/generator.go
@@ -19,8 +19,10 @@ import (
 	"flag"
 	"os"
 	"reflect"
+	"strconv"
 
 	"github.com/gardener/gardener-extension-provider-azure/pkg/apis/azure/v1alpha1"
+	"github.com/go-logr/logr"
 
 	"github.com/gardener/gardener/extensions/test/tm/generator"
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


test/tm/generator.go
```diff
@@ -29,20 +31,41 @@ import (
...
+	cfg    *GeneratorConfig
+	logger logr.Logger
 )
 
+func addFlags() {
+	cfg = &GeneratorConfig{}
+	flag.StringVar(&cfg.infrastructureProviderConfigPath, "infrastructure-provider-config-filepath", "", "filepath to the provider specific infrastructure config")
+	flag.StringVar(&cfg.controlplaneProviderConfigPath, "controlplane-provider-config-filepath", "", "filepath to the provider specific controlplane config")
+
+	flag.StringVar(&cfg.networkVnetCidr, "network-vnet-cidr", "10.250.0.0/16", "vnet network cidr")
```
dguendisch, line 58:
```suggestion
	flag.StringVar(&cfg.networkVnetCidr, "network-vnet-cidr", "", "vnet network cidr")
```

I think `validate()` already sets the default, so probably remove it here.

dguendisch (approved):
/lgtm

schrodit (approved):

rfranzke (approved):

PR #191, opened 2020-11-16
Error wrapping for UpdateMachineImagesStatus function

rfranzke:

/area logging
/kind enhancement
/priority normal
/platform azure

Introduce error wrapping for `UpdateMachineImagesStatus` function.

Part of https://github.com/gardener/gardener/issues/3119

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/controller/worker/machine_images.go
+++ b/pkg/controller/worker/machine_images.go
@@ -29,18 +29,22 @@ import (
 func (w *workerDelegate) UpdateMachineImagesStatus(ctx context.Context) error {
 	if w.machineImages == nil {
 		if err := w.generateMachineConfig(ctx); err != nil {
-			return err
+			return errors.Wrapf(err, "unable to generate the machine config")
 		}
 	}
 
 	// Decode the current worker provider status.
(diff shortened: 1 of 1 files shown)


kon-angelo:
I know that currently there are mixed approaches used in the code about error wrapping, but how would you prefer 
`return fmt.Errorf("<error message>: %w", err)` (eventually converting cleaning the rest of the places too). 

It is subtle but I am not sure we need the stack trace of the Wrapf instead of a simple wrap using the std library.I would also argue that we instead of `%w` we could simply `%v` since we do not use the wrapped source error in any way upstream the stack.

rfranzke:
@stoyanr can you please comment on that one? You probably have an opinion about it as you initially introduced the `errors.Wrapf()`.

rfranzke:
ping @stoyanr

kon-angelo (approved):
@rfranzke the differences are just minor (the extra dependency compared to `fmt`+ the extra stack trace compared to my suggestion) and even if my proposal was accepted here the usage in the code is mixed at the moment.  

I was partially interested in seeing other people opinions for setting a more standard approach.

/lgtm

PR #210, opened 2020-12-02
Validation: Allow to resize the vnet cidr

dkistner:

/area open-source
/kind enhancement
/kind bug
/priority normal
/platform azure

Allow to resize the vnet cidr of a Shoot cluster as this is possible on Azure and handy for some migration scenarios.
Fix also an issue where the old and new shoots are never compared in an update case.
Restructure the vnet validation code.

**Release note**:

```feature user
It is now possible to resize the cidr range of a managed vnet for Azure shoot clusters.
```

```bugfix developer
An issue is fixed where the old and new shoots are never compared to each other in an update case.
```

### DIFF
--- a/pkg/apis/azure/validation/infrastructure.go
+++ b/pkg/apis/azure/validation/infrastructure.go
@@ -57,35 +57,13 @@ func ValidateInfrastructureConfig(infra *apisazure.InfrastructureConfig, nodesCI
 
 	networksPath := fldPath.Child("networks")
 
+	// Validate workers subnet cidr
 	workerCIDR := cidrvalidation.NewCIDR(infra.Networks.Workers, networksPath.Child("workers"))
-
 	allErrs = append(allErrs, cidrvalidation.ValidateCIDRParse(workerCIDR)...)
 	allErrs = append(allErrs, cidrvalidation.ValidateCIDRIsCanonical(networksPath.Child("workers"), infra.Networks.Workers)...)
 
-	if (infra.Networks.VNet.Name != nil && infra.Networks.VNet.ResourceGroup == nil) || (infra.Networks.VNet.Name == nil && infra.Networks.VNet.ResourceGroup != nil) {
-		allErrs = append(allErrs, field.Invalid(networksPath.Child("vnet"), infra.Networks.VNet, "specifying an existing vnet name require a vnet name and vnet resource group"))
-	} else if infra.Networks.VNet.Name != nil && infra.Networks.VNet.ResourceGroup != nil {
-		if infra.Networks.VNet.CIDR != nil {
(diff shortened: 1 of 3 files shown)


pkg/apis/azure/validation/infrastructure.go
```diff
@@ -114,17 +92,83 @@ func ValidateInfrastructureConfig(infra *apisazure.InfrastructureConfig, nodesCI
...
 	}
 
 	return allErrs
 }
+
+func validateVnetConfigUpdate(oldNeworkConfig, newNetworkConfig *apisazure.NetworkConfig, networkConfigPath *field.Path) field.ErrorList {
+	var allErrs = field.ErrorList{}
+
+	if newNetworkConfig.VNet.CIDR == nil && newNetworkConfig.VNet.Name == nil && newNetworkConfig.VNet.ResourceGroup == nil {
+		return append(allErrs, field.Invalid(networkConfigPath.Child("vnet"), newNetworkConfig.VNet, "a vnet cidr or vnet name and resource group need to be specified"))
```
kon-angelo, line 150: Since this is not an old vs new comparison, wouldn't that be covered in the normal validation `validateVnetConfig` ?

dkistner: Hmm no. It is valid to create a shoot without passing a vnet config at all. Then it will be defaulted. But it shouldn't be valid to remove the vnet config afterwards.

dkistner: Since we apply a default value in the controller itself here: https://github.com/gardener/gardener-extension-provider-azure/blob/master/pkg/internal/infrastructure/terraform.go#L118-L119

I agree it does not make sense to add this update validation yet. We should consider to add this defaulting not in the controller. Will open an issue for that and remove the validation for now.

kon-angelo:
I think that as it is setup right now, the update test will basically allow any update in vnet configuration if the `oldconfig.networks.Vnet` is blank on shoot creation.

Possible test case:
```go
			It("should it forbid adding external config??", func() {
				infrastructureConfig.Networks.VNet = apisazure.VNet{}

				newInfrastructureConfig := infrastructureConfig.DeepCopy()
				newInfrastructureConfig.Networks.VNet.Name = pointer.StringPtr("modified")
				newInfrastructureConfig.Networks.VNet.ResourceGroup = pointer.StringPtr("modified")

				errorList := ValidateInfrastructureConfigUpdate(infrastructureConfig, newInfrastructureConfig, fldPath)
				Expect(errorList).ToNot(BeEmpty())
			})
		})
```

This test fails as the `errorList` is empty, but I think that it should prohibited this modification right ?

dkistner:
Yes makes sense to validate this case as well. Done.
@kon-angelo PTAL :)

kon-angelo (approved):
/lgtm

PR #213, opened 2020-12-04
Fix: Terraforrm destroy after azurerm v2 upgrade

dkistner:

/area open-source
/kind bug
/priority normal
/platform azure

Infrastructures which are created with tf azurerm v1 and get deleted without getting reconciled first with tf azurerm v2 will stuck as Terraform tries to destroy the Infrastructure based on a tf azurerm v1 config. Therefore we need to render the new tf azurerm config before deletion.

Credits to @ialidzhikov 

1. Create an Infra with provider-azure e.g `v1.14.1` (does not contain a Terraformer with azurerm v2)
2. Stop provider-azure extension (to avoid reconciliation of the Infra) and delete the Infra
3. Checkout this changes and start the provider-azure extension, it should now delete the Infra properly 

**Release note**:

```bugfix user
A bug has been fixed which prevented the deletion of Azure Infrastructures in relation to the upgrade of the azurerm v2 Terraform provider.
```

/invite @ialidzhikov 
/invite @kon-angelo


dkistner:
fixing the ci

ialidzhikov (approved):
/lgtm

Issue #3, opened 2020-01-10
Adopt CSI for Azure extension beginning with Kubernetes 1.21


rfranzke:
As of Kubernetes 1.18 the in-tree Azure Disk volume driver is deprecated in favour of CSI. We should adopt CSI for 1.18+ clusters (where the Azure Disk migration will be promoted to beta as well)

See also: https://github.com/kubernetes/kubernetes/pull/85231 and https://kubernetes.io/blog/2019/12/09/kubernetes-1-17-feature-csi-migration-beta/

rfranzke:
The CSI migration is beta with 1.17, though, the CSI plugins for Azure are not yet ready to cater with Gardener needs. We will have to contribute there before we can migrate to CSI. Hence, postponing this issue until 1.19.

rfranzke:
I opened https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/354 for the Azure Disk CSI plugin. After it's merged I'll use the same approach to enhance the [Azure File CSI plugin](https://github.com/kubernetes-sigs/azurefile-csi-driver).

rfranzke:
Turns out that what I described with https://github.com/gardener/gardener-extension-provider-azure/issues/3#issuecomment-610922049 was actually a misconfiguration during my tests - the Azure Disk CSI plugin already works for our scenario as it disables the metadata service in the controller service of the CSI driver. Hence https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/354 is closed again for now.

However, it seems that the node service of the CSI driver also requires the client id/secret credentials in order to talk to the Azure API and find out information about certain disks. This would mean that we would have to expose the credentials in the shoot cluster which is undesired. I'm investigating further...

rfranzke:
I checked a little bit more, and it seems that the [`checkDiskExists()`](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/v0.6.0/pkg/azuredisk/nodeserver.go#L331-L333) function in the `NodeGetVolumeStats` function is the only place in the node service where the credentials are needed right now.
The Azure CSI maintainers offered to look into whether it is possible to get rid of this check. Afterwards it should be possible to start the CSI driver without client id/secret.

rfranzke:
[...]

PR #215, opened 2020-12-05
Postpone CSI migration to Kubernetes v1.21

ialidzhikov:
/kind enhancement
/priority normal
/platform azure

Ref https://github.com/gardener/gardener-extension-provider-azure/issues/3#issuecomment-737330926

**Release note**:

```other user
CSI controllers / drivers are now deployed by default if the Kubernetes version is v1.21 or greater (not v1.20 as before). The reason for postponement is that the corresponding upstream feature `CSIMigrationAzureFile` is still alpha in Kubernetes v1.20 and its promotion to beta is pushed back to v1.21.
```

### DIFF
--- a/docs/usage-as-end-user.md
+++ b/docs/usage-as-end-user.md
@@ -244,10 +244,10 @@ spec:
 
 ## CSI volume provisioners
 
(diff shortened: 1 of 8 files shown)


vpnachev (approved):
/lgtm

dkistner (approved):
/lgtm

Issue #101, opened 2020-05-30
Properly Enable Accelerated Networking for AvSet Clusters


vpnachev:
**How to categorize this issue?**

/area performance
/area  quality
/kind enhancement
/priority normal
/platform azure

**What would you like to be added**:
Recently Accelerated Networking was enabled with #65. The decision to enable it or not was based only on the VM type and OS, but this caused troubles with AvSet clusters - basically the migration to VMs with enabled Accelerated Networking is requiring all non-enabled VMs to be deleted before new ones are created. If this is not done, VM creation is failing with `ExistingAvailabilitySetWasNotDeployedOnAcceleratedNetworkingEnabledCluster`, thus we disabled Accelerated Networking with #100 for all AvSet based shoot clusters. 

Once https://github.com/gardener/gardener-extension-provider-azure/issues/61 is implemented, we can use AvSet per machine class which will allow us to properly enable Accelerated Networking for non-zoned clusters, again.

**Why is this needed**:
See #55

kon-angelo:
AvSet not supported anymore

PR #217, opened 2020-12-07
Introduce Azure VMSS orchestration mode VM (VMO) based Shoot clusters 

### LINKED ISSUE #61: Multiple (Availability)Sets
Gardener currently deploy for non zoned clusters only one Availability Set for all machines in the cluster.
This approach has several drawbacks, e.g. not possible to combine different machine type skus, no option to move to newer machine type skus, stability and performance of update operations etc.

dkistner:

/kind enhancement
/priority normal
/platform azure

Add support for Azure virtual machine scaleset orchestration mode vm (VMO) based Shoot clusters.
Each worker pool will get its own VMO instead its of a shared resource like for AvailabilitySets based clusters.

The Shoot cluster need to be non-zoned and have the annotation `alpha.azure.provider.extensions.gardener.cloud/vmo=true` to enable VMO which we currently treat as a preview feature.

**Details:**
- In case the corresponding fault domain count in the CloudProfile changes then the VMO will be replaced. This will lead to a rolling update of the machines in each worker pool.
- Orphan Gardener managed VMOs will be removed during the Worker reconciliation. So it is safe to be deployed into existing resource group without leaking resources.
- VMO based clusters will make use of the Standard SKU Loadbalancer.
- The NatGateway will be available for VMO clusters. This is a difference to AvailabilitySet based clusters which does not support NatGateway.
- Accelerated network will work with VMO clusters. We can't enable it for AvailabilitySet based clusters as all machines assigned to it need to be compatible/running with accelerated network support which is we can't archive for existing clusters with a central AvailabilitySet, see https://github.com/gardener/gardener-extension-provider-azure/issues/101

**More details:**
- The Azure SDK is updated to support compute client in version `2020-06-30` (min `2019-12-01` is required to use VMO)
- VMO clusters will use the `vmss` version of the vm-controller in the Azure cloud-controller-manager (activated in the cloud-provider config via `vmType=vmss`).
- A vmss client has been added to the client factory.
- Added mocks for the client factory and the vmss client.
- The Worker controller tests have been restructured partially to allow more reuse for the machine dependency tests.
- The machineclass chart structure to configure AvailabilitySets has been aligned to the way to configure VMOs.

Intend to fix #61

**Release note**:

```feature enduser
Gardener is now supporting Azure clusters based on Virtual MachineScale Orchestration Mode Vm (VMO) as a preview feature. 
VMO should be perspectively a replacement for AvailabilitySet based clusters. You need to make sure that your subscription is eligible for VMO usage. You can find more details here: https://github.com/gardener/gardener-extension-provider-azure/blob/master/docs/usage-as-end-user.md#preview-shoot-clusters-with-vmss-orchestration-mode-vm-vmo
```

/invite @kon-angelo


pkg/controller/worker/machine_dependencies_test.go
```diff
@@ -1,35 +1,318 @@
...
+			}
 		})
-	})
 
-	Context("#CleanupMachineDependencies", func() {
-		It("should return no error", func() {
-			err := workerDelegate.CleanupMachineDependencies(context.TODO())
-			Expect(err).NotTo(HaveOccurred())
+		Context("#DeployMachineDependencies", func() {
+			It("should deploy no vmo dependency as it not required", func() {
```
ialidzhikov, line 113:
```suggestion
			It("should deploy no vmo dependency as it is not required", func() {
```

pkg/controller/worker/machine_dependencies_test.go
```diff
@@ -1,35 +1,318 @@
...
+
+				workerStatus := decodeWorkerProviderStatus(w)
+				Expect(workerStatus.VmoDependencies).To(ContainElements(MatchFields(IgnoreExtras, Fields{
+					"ID":       Equal(vmoDependency.ID),
+					"Name":     Equal(vmoDependency.Name),
+					"PoolName": Equal(vmoDependency.PoolName),
+				})))
+			})
+
+			It("should cleanup a vmo dependency as corresponding worker pool does not exists anymore", func() {
```
ialidzhikov, line 223:
```suggestion
			It("should cleanup a vmo dependency as corresponding worker pool does not exist anymore", func() {
```

pkg/controller/worker/machines.go
```diff
@@ -23,6 +23,7 @@ import (
 	"strings"
 
 	azureapi "github.com/gardener/gardener-extension-provider-azure/pkg/apis/azure"
+	"github.com/gardener/gardener-extension-provider-azure/pkg/apis/azure/helper"
```
ialidzhikov, line 26: Already imported in the line below.

```suggestion
```

ialidzhikov (commented):
Minor nits from my initial "go through".

ialidzhikov (changes requested):
@dkistner , the infrastructure test needs also to be adapted. Ref https://tm.gardener.cloud/testrun/default/e2e-sdh4d

Other than that I played locally with the change and it worked well for lifecycle operations for Shoot with enabled VMO.

dkistner:
Awesome. Thanks @ialidzhikov for the review.
I will address the changes in the new year :)

pkg/azure/client/factory.go
```diff
@@ -19,6 +19,7 @@ import (
 
 	"github.com/gardener/gardener-extension-provider-azure/pkg/internal"
 
+	azurecompute "github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-12-01/compute"
```
ialidzhikov, line 22: Currently the provider imports `"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-12-01/compute"` and the infrastructure test imports `"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2020-06-01/compute"` which causes both versions of the package to be vendored. So we should stick to one of the versions and use it across the project.

After applying the following change on `test/integration/infrastructure/infrastructure_test.go`

```diff
-	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2020-06-01/compute"
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-12-01/compute"
```

```
$ make revendor

$ git diff --shortstat
 36 files changed, 1 insertion(+), 33388 deletions(-)
```

=> `"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2020-06-01/compute"` becomes obsolete

ialidzhikov (commented):
Thank you, @dkistner !
Adding one last comment that I missed to add in the previous review.

pkg/azure/client/factory.go
kon-angelo: I would do it the other way around though to get the newer version

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -503,13 +495,13 @@ func getConfigChartValues(
...
-			return nil, errors.Wrapf(err, "could not determine availability set for purpose 'nodes'")
-		}
-		values["availabilitySetName"] = nodesAvailabilitySet.Name
+	primaryAvailabilitySet, err := helper.FindAvailabilitySetByPurpose(infraStatus.AvailabilitySets, apisazure.PurposeNodes)
+	if err == nil {
+		values["availabilitySetName"] = primaryAvailabilitySet.Name
+	}
+
+	if helper.IsVMORequired(infraStatus) {
+		values["vmType"] = "vmss"
```
kon-angelo, line 504: Maybe 
```go
	if primaryAvailabilitySet, err := helper.FindAvailabilitySetByPurpose(infraStatus.AvailabilitySets, apisazure.PurposeNodes); err == nil {
		values["availabilitySetName"] = primaryAvailabilitySet.Name
	} else if helper.IsVMORequired(infraStatus) {
		values["vmType"] = "vmss"
	}
```
I think the `else` does add something here, by highlighting that vmos and AVs are mutually exclusive. I would maybe consider placing the `isVMORequired` on top and the AVs in the else clause, because the condition inside the func do take into account the AVs.

pkg/internal/infrastructure/terraform.go
```diff
@@ -440,6 +462,43 @@ func findDomainCounts(cluster *controller.Cluster, infra *extensionsv1alpha1.Inf
...
+	// If the infrastructureStatus already exists that mean the Infrastucture is already created.
+	infrastructureStatus, err := helper.InfrastructureStatusFromInfrastructure(infra)
+	if err != nil {
+		return false, err
+	}
+
+	if len(infrastructureStatus.AvailabilitySets) > 0 {
+		if _, err := helper.FindAvailabilitySetByPurpose(infrastructureStatus.AvailabilitySets, api.PurposeNodes); err == nil {
+			if hasVmoAnnotation {
+				return false, errors.New("Can't use vmss orcestration mode VM as this cluster already used an availability set")
```
kon-angelo, line 493: + `orchestration` typo
+ small case for the error (`cannot`)

pkg/azure/client/factory.go
ialidzhikov: Yep, sure, makes sense.

dkistner: I will even go a step further and update the compute sdk to the most recent version `2020-06-30`.

pkg/internal/infrastructure/terraform.go
dkistner: done

pkg/controller/controlplane/valuesprovider.go
dkistner: I moved into a seperate function to only append the respective key if a condition match and return then.

pkg/azure/client/factory.go
dkistner: done

dkistner:
Thanks for the feedback. I have addressed it.
Also I added infrastructure integration tests for VMO as they are missing before.

/test

ialidzhikov (approved):
/lgtm

rfranzke:
@dkistner @kon-angelo Can this PR be merged?

kon-angelo (approved):
/lgtm too

Let's wait a final confirmation from @dkistner.

dkistner:
Thanks for confirming. Yes lets go on and merge :)

danielfoehrKn:
Very nice PR!

I could not find the place in the MCM azure machine driver where the VM is created with a reference to the ScaleSet. Can you by any chance point me to it? Thanks!

dkistner:
Hi Daniel, see here: https://github.com/gardener/machine-controller-manager/blob/master/pkg/driver/driver_azure.go#L191-L194

Chokiis82 (commented):
- [ ] ``

PR #224, opened 2020-12-11
Fix deleting controlplane with purpose exposure

stoyanr:

/area control-plane
/kind bug
/priority critical
/platform azure

Changes the logic in the controlplane actuator so that remedy controller resources are deleted only for `normal` controlplanes, but not for `exposure` ones.

Fixes #

There is a step in the reconcile flow that deletes the `exposure` controlplane only if the `APIServerSNI` feature gate is enabled. Without this fix, this step fails in a hibernated cluster since the remedy controller resources are also deleted, but the remedy controller itself is scaled to zero and can't reconcile the deletion.

**Release note**:

```other operator
Fixed a bug that caused reconciling hibernated shoots to fail if the `APIServerSNI` feature gate is enabled.
```

### DIFF
--- a/pkg/controller/controlplane/actuator.go
+++ b/pkg/controller/controlplane/actuator.go
@@ -69,9 +69,11 @@ func (a *actuator) Delete(
 	cp *extensionsv1alpha1.ControlPlane,
 	cluster *extensionscontroller.Cluster,
 ) error {
-	// Delete all remaining remedy controller resources
-	if err := a.deleteRemedyControllerResources(ctx, cp); err != nil {
-		return err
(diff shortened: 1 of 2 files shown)


rfranzke (approved):
/lgtm
/needs second-opinion
/invite @dkistner @kon-angelo

ialidzhikov (approved):
/lgtm

danielfoehrKn (approved):
/lgtm  Thanks for finding!

dkistner (approved):
/lgtm

rfranzke:
Nice, thanks for merging. Can we have a hotfix release with this PR?

dkistner:
yep will do it

PR #258, opened 2021-03-07
add minAllow for VPA of provider-azure extension itself

MartinWeindel:

/area control-plane
/area auto-scaling
/area robustness
/kind enhancement
/platform azure

The gardener-extension-provider-azure pods on one seed have been in `CrashLoopBackOff` (OOMKilled) because the VPA recommendations for memory were ridiculous low (10 MB). Therefore `minAllow` settings for VPA should be activated.

Fixes #

**Release note**:

```other operator
provider-azure does now specify minAllowed values by default for the extension controller to prevent too low recommendations from VPA that lead to OOMK.
```


ialidzhikov:
@MartinWeindel , we talked about this PR in the quality responsible sync and the last agreement was that these values might be gardener landscape specific and we should rather specify them per gardener landscape (it is possible to pass these values via the `ControllerRegistration` spec).
@rfranzke , correct me if I got it wrong.

ialidzhikov:
> @MartinWeindel , we talked about this PR in the quality responsible sync and the last agreement was that these values might be gardener landscape specific and we should rather specify them per gardener landscape (it is possible to pass these values via the ControllerRegistration spec).

So rethinking about the above statement, I see that it is still reasonable to have some bare minimum specified for minAllowed to prevent ridiculous VPA recommendation. In this way we would also prevent the issue described in https://github.com/gardener/gardener/pull/2164 (VPA recommending less memory than the size of the runc binary (about 10Mb)).
I guess we don't want to specify and maintain minAllowed values for each and every extension and for each and every landscape. If custom minAllowed are needed for a landscape, then it can be customized.

charts/gardener-extension-provider-azure/values.yaml
```diff
@@ -10,10 +10,10 @@ maxSurge: 50%
...
 vpa:
   enabled: true
-# resourcePolicy:
-#   minAllowed:
-#     cpu: 100m
-#     memory: 128Mi
+  resourcePolicy:
+    minAllowed:
+      cpu: 10m
+      memory: 120Mi
```
ialidzhikov, line 16: I would rather stick to lower minAllowed wrt the memory. For example the values from os-gardenlinux extension - https://github.com/gardener/gardener-extension-os-gardenlinux/blob/94ea26e9be17b375e70e24bed432e78197d7a330/charts/gardener-extension-os-gardenlinux/values.yaml#L12-L13

```suggestion
      cpu: 50m
      memory: 64Mi
```

ialidzhikov: `make generate` is needed :(

ialidzhikov (approved):
/lgtm

PR #266, opened 2021-03-23
Use separate ManagedResource for ControlPlane CRDs

ialidzhikov:
/kind enhancement
/platform azure

Part of https://github.com/gardener/gardener/issues/2227

As the CSI is still not enabled for provider-azure (planned for K8s v1.21), we can move the volumesnapshot CRDs to separate ManagedResource in a backwards-compatible way (for the rest of the provider extensions we will have to migrate the CRDs from one ManagedResource to another one).

**Release note**:

```other operator
provider-azure is now using a separate ManagedResource for ControlPlane CRDs that are installed in the Shoot cluster to separate the deletion of CRDs from the deletion of the RBAC for controller leader election.
```


rfranzke (approved):
/lgtm

dkistner (approved):
/lgtm

PR #271, opened 2021-03-29
Deserialize custom resources with EnableStrict

### LINKED ISSUE #270: Deserialize extension resources (Infrastructure, etc.) using strict deseriazliation, including during validation

/area control-plane
/kind enhancement
/priority 3
/platform azure

Extension resources such as `Infrastructure` should be deserialized using the so-called "strict mode" that was introduced recently. In this mode, deserialization fails if there is any data that is not part of the API definition.

**Why is this needed**:

We recently had the case where the infrastructure creation failed because the user did not indent `natGateway` properly in an infrastructure resource :

```yaml
kind: Shoot
apiVersion: core.gardener.cloud/v1beta1
spec:
  provider:
    type: azure
    infrastructureConfig:
      kind: InfrastructureConfig
      networks:
        ...
      natGateway: # <--- nat gateway should be under networks
        ...
```

Changing the deserialization as described would enable us to detect and report such mistakes earlier, e.g. when the Shoot resource is created / updated, rather than much later when the infrastructure creation fails.

We would like to introduce this first for azure, later for other providers as well.

stoyanr:

/area control-plane
/kind enhancement
/priority 3
/platform azure

Adds the `EnableStrict` option to all places where a codec factory is creation, to ensure that custom resources are always deserialized in "strict" mode. This means that resources with fields that are not allowed by the API schema will be rejected by validation and would not be processed by the controllers as well.

Fixes #270

**Release note**:

```breaking user
Extension resources (`Infrastructure`, `ControlPlane`, etc.) are now deserialized in "strict" mode, including during validation by the validating webhook. This means that resources with fields that are not allowed by the API schema will be rejected by validation. Creating new shoots containing such resources will not be possible, and updating existing shoots will fail with an appropriate error until you manually update the shoot to make sure any extension resources contained in it are valid.
```

### DIFF
--- a/pkg/admission/validator/shoot.go
+++ b/pkg/admission/validator/shoot.go
@@ -21,7 +21,6 @@ import (
 
 	extensionswebhook "github.com/gardener/gardener/extensions/pkg/webhook"
 	"github.com/gardener/gardener/pkg/apis/core"
-	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/util/validation/field"
@@ -44,8 +43,8 @@ var (
 
 // shoot validates shoots
 type shoot struct {
-	decoder runtime.Decoder
-	Logger  logr.Logger
+	decoder        runtime.Decoder
+	lenientDecoder runtime.Decoder
 }
 
 // NewShootValidator returns a new instance of a shoot validator.
@@ -55,7 +54,8 @@ func NewShootValidator() extensionswebhook.Validator {
 
 // InjectScheme injects the given scheme into the validator.
 func (s *shoot) InjectScheme(scheme *runtime.Scheme) error {
-	s.decoder = serializer.NewCodecFactory(scheme).UniversalDecoder()
+	s.decoder = serializer.NewCodecFactory(scheme, serializer.EnableStrict).UniversalDecoder()
+	s.lenientDecoder = serializer.NewCodecFactory(scheme).UniversalDecoder()
 	return nil
 }
 
@@ -121,7 +121,7 @@ func (s *shoot) validateUpdate(oldShoot, shoot *core.Shoot) error {
 	if oldShoot.Spec.Provider.InfrastructureConfig == nil {
 		return field.InternalError(infraConfigPath, fmt.Errorf("InfrastructureConfig is not available on old shoot"))
 	}
-	oldInfraConfig, err := checkAndDecodeInfrastructureConfig(s.decoder, oldShoot.Spec.Provider.InfrastructureConfig, infraConfigPath)
+	oldInfraConfig, err := checkAndDecodeInfrastructureConfig(s.lenientDecoder, oldShoot.Spec.Provider.InfrastructureConfig, infraConfigPath)
 	if err != nil {
 		return err
 	}
--- a/pkg/admission/validator/cloudprofile.go
+++ b/pkg/admission/validator/cloudprofile.go
@@ -39,7 +39,7 @@ type cloudProfile struct {
 
 // InjectScheme injects the given scheme into the validator.
 func (cp *cloudProfile) InjectScheme(scheme *runtime.Scheme) error {
-	cp.decoder = serializer.NewCodecFactory(scheme).UniversalDecoder()
+	cp.decoder = serializer.NewCodecFactory(scheme, serializer.EnableStrict).UniversalDecoder()
 	return nil
 }
 
--- a/pkg/apis/azure/helper/scheme.go
+++ b/pkg/apis/azure/helper/scheme.go
@@ -43,7 +43,7 @@ func init() {
 	Scheme = runtime.NewScheme()
 	utilruntime.Must(install.AddToScheme(Scheme))
 
-	decoder = serializer.NewCodecFactory(Scheme).UniversalDecoder()
+	decoder = serializer.NewCodecFactory(Scheme, serializer.EnableStrict).UniversalDecoder()
 }
 
 // InfrastructureConfigFromInfrastructure extracts the InfrastructureConfig from the
--- a/pkg/controller/worker/utils_test.go
+++ b/pkg/controller/worker/utils_test.go
@@ -48,7 +48,7 @@ func wrapNewWorkerDelegate(client *mockclient.MockClient, seedChartApplier *mock
 	_ = apiazure.AddToScheme(scheme)
 	_ = v1alpha1.AddToScheme(scheme)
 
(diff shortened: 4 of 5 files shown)


vpnachev:
As an operator, I would like to have the option to disable this check for at least several releases. Can you make it configurable? It can still be enabled by default.

rfranzke:
+1, good point @vpnachev, I think this makes sense.

stoyanr:
@rfranzke @vpnachev Are you sure this is really a must, and if yes, why? Do you expect that users are using a lot of shoots with wrongly formatted extensions, or are you afraid of something else?

Although it may not seem much of a change, please have in mind that it should be done for every single provider extension (and maybe for others as well?), so the effort is multiplied by a factor of ~10.

In case it has to be configurable, should it be configurable for both the extension controller and the admission webhook executables, and how exactly? While the controller is configurable via a config file (and it would be natural to add it there), the admission webhook doesn't have such a config file, so it should either be added (making the change even bigger), or a command line switch should be used (making it inconsistent with the controller). Or, both could use command line switches, but for the controller it would be rather weird.

Unless you are 100% sure this is a must, I would save this entire complication.

rfranzke:
OK, that's true, it would mean a lot of effort on our side given that it requires configurability in both components and for all provider extensions. Given this, I rather also tend to not making it configurable and rather argue that such resources are already broken/invalid in the first place. Landscape operators still have the option to announce that this strict validation will be enabled before they deploy the next release of the extensions.
I think in k/k it's not configurable neither (of course, k/k is somewhat different compared to Gardener because end-users control the version (+ resulting behavioral changes) themselves).

stoyanr:
@rfranzke Thanks, in this case I would expect a /lgtm, or do you have also other concerns?

rfranzke (dismissed):
Nope, looks good, thank you!

/lgtm
/invite @vpnachev for
/second-opinion

vpnachev:
Will this be properly categorized as user error in the shoot status?
I do not see release not of type `breaking`.

stoyanr:
@vpnachev 
> Will this be properly categorized as user error in the shoot status?

Since the error will first popup in the admission webhook, the shoot will not be created at all. 

> I do not see release note of type `breaking`.

Could you please be more concrete what do you think should be done?

vpnachev:
> > Will this be properly categorized as user error in the shoot status?
> 
> Since the error will first popup in the admission webhook, the shoot will not be created at all.

@stoyanr what about update/patch requests?



> > I do not see release note of type `breaking`.
> 
> Could you please be more concrete what do you think should be done?

@stoyanr the release note that you have added to this PR does not reflect the real change and that you are about to potentially brake users. 
If I am not a Gardener developer, but just a Gardener Landscape user, I will never ever go through any PR in the various components used in a landscape, instead I would read only the release notes. So please, add a `breaking user` release note and describe as precisely as possible what this PR is about to change and how/when some users can be affected.

ialidzhikov:
> I do not see release note of type `breaking`.

Yes, valid point. Sorry for joining late, but what is the update/migration strategy wrt to this change? IIUC this is a braking change and if a Shoot has invalid infrastructureConfig, then then admission will disallow any updates on such Shoot? If that is the case how we fix these Shoots and allow gardenlet and any other component to update them?

stoyanr:
> ... how we fix these Shoots and allow gardenlet and any other component to update them?

The admission will only disallow updates to shoots if they still contain invalid extension resources. Fixing them is as simple as making sure the extension resources are valid. A small number of existing user shoots may be affected (I don't think users habitually created such shoots), in which case the users themselves will have to fix them.

> ... what about update/patch requests?

I will check this.

Finally, a request: if you still have concerns, let's discuss them face to face.

vpnachev:
> let's discuss them face to face.

Definitely no. Just 6 months later nobody will remember what we have been discussing f2f.
It is much better the reasoning for the decision to be in the PR.

stoyanr:
@vpnachev Let's be realistic. We aren't doing that for most other PRs as well, it's simply far, far too much. I will take care to update the PR with whatever changed and the reasoning for it. Meanwhile, let's try to be somewhat more efficient and understand each other better.

stoyanr:
> Without having checked the implementation, my assumption is that status updates to the shoots are not even calling this admission webhook, so the gardenlet would be fine and can continue to operate, I assume (is this correct @stoyanr?). It's just that the spec cannot be changed unless the configurations are fixed, right?

Unfortunately, right now this is not the case ... Updates break even if the old shoot is invalid :-( This should be fixed of course. Regarding the status, I am not yet sure (simply haven't got that far). Let me properly test and fix this, hopefully I can make it work exactly as you outlined.

stoyanr:
I fixed the update issue and can confirm it works exactly as @rfranzke outlined above - the processing of the Shoot fails with an error in the infrastructure controller and it's properly reflected in the status. This error is not yet categorized as an user error (e.g. `ERR_CONFIGURATION_PROBLEM`). Since the categorization function is in g/g, it follows I should file a PR for g/g for it, there is nothing to do here, correct?

Until a newer g/g is vendored, it will work as expected except the categorization, which should not be an issue if we take care to vendor a newer g/g in all extensions before their next release?

stoyanr:
The PR in g/g for updating the error categorizaton is https://github.com/gardener/gardener/pull/3798

pkg/admission/validator/shoot.go
```diff
@@ -44,8 +44,9 @@ var (
 
 // shoot validates shoots
 type shoot struct {
-	decoder runtime.Decoder
-	Logger  logr.Logger
+	decoder        runtime.Decoder
+	lenientDecoder runtime.Decoder
+	Logger         logr.Logger
```
rfranzke, line 49: Is this Logger used somewhere? Why is it exported?

stoyanr: No, it's not, perhaps some leftover. It has nothing to do with my change, but I will delete it anyway.

stoyanr: Done

rfranzke: Ah, I missed that it wasn't even added by you. Anyways, thanks for cleaning up!

rfranzke (approved):
/lgtm

dkistner (approved):
Sorry for joining late. Following the discussions it is looking good to me.

How to proceed here? Should we wait until g/g with the related pr is released? 
Or can we merge it and incorporate it in the next release (which will happen quite soon)?

stoyanr:
@dkistner I am also not really sure. Since the g/g change is already merged, I could vendor a newer g/g with it, but I am not really sure if this effort is needed. @rfranzke What would you advise?

rfranzke:
We agreed to only vendor released versions of g/g (at least when we want to release a new extension version). Anyways, I think there is no strong need to do anything now. This error code mapping is just to ease the operator's lives, so it can also come a little later.

stoyanr:
OK, I will then proceed with making the same change for other extensions.

dkistner:
Alright then we can bring that in and incorporate it in our next release.

PR #276, opened 2021-04-09
Fix vmo cluster deletion if the worker status is empty

dkistner:
or to be more precise if the worker status does not no contain a corresponding vmo dependency.

/kind bug
/platform azure

If the corresponding vmss orcmode vm/flexmode (vmo) Azure resource (e.g. quota exceeded) can't be created for an Azure cluster, then the workerstatus of the pool will not contain a dependency reference for the vmo. If a user then decided to delete the cluster, the machinedeployment/class will be regenerated during the deletion flow. During this regeneration the vmo dependency reference in the workerstatus will be accessed. If this is not present the provider-azure extension will panic due to a nilpointer exception.

**Release note**:

```bugfix user
A bug is fixed which blocked the deletion of vmo based Azure cluster in case the vmo resource on Azure cannot be created.
```

/invite @kon-angelo 
/cc @timebertt

### DIFF
--- a/pkg/controller/worker/machineset_vmo.go
+++ b/pkg/controller/worker/machineset_vmo.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
 
 	azureapi "github.com/gardener/gardener-extension-provider-azure/pkg/apis/azure"
 	azureapihelper "github.com/gardener/gardener-extension-provider-azure/pkg/apis/azure/helper"
@@ -96,11 +97,7 @@ func (w *workerDelegate) reconcileVMO(ctx context.Context, client azureclient.Vm
 		return newVMO, nil
 	}
 
-	return &azureapi.VmoDependency{
-		ID:       *vmo.ID,
-		Name:     *vmo.Name,
-		PoolName: workerPoolName,
-	}, nil
+	return generateVmoDependency(vmo, workerPoolName), nil
 }
 
 func (w *workerDelegate) cleanupVmoDependencies(ctx context.Context, infrastructureStatus *azureapi.InfrastructureStatus, workerProviderStatus *azureapi.WorkerStatus) ([]azureapi.VmoDependency, error) {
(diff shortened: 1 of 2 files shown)


timebertt:
Thanks for taking care! :)

pkg/controller/worker/machineset_vmo.go
```diff
@@ -172,6 +169,75 @@ func cleanupOrphanVMODependencies(ctx context.Context, client azureclient.Vmss,
...
+	// First: Lookup the vmo dependency for the worker pool in the worker status.
+	var dependencyInStatus *azureapi.VmoDependency
+	for _, dep := range workerStatus.VmoDependencies {
+		if dep.PoolName != workerPoolName {
+			continue
+		}
+		if dependencyInStatus != nil {
+			return nil, fmt.Errorf("found more then one vmo dependencies for workerpool %s in the worker provider status", workerPoolName)
+		}
+		runVar := dep
```
kon-angelo, line 186: If I understand correctly this is done because `dep` is reassigned in every loop iteration and you need to snapshot the value. In this case the purpose of `runVar` is not very clear  because the name is not clear and you would need to be aware of the loop pitfall to understand why it is there. 

Is maybe something like this more clear? 
```go
	for _, d := range workerStatus.VmoDependencies {
                dep := d
		if dep.PoolName != workerPoolName {
			continue
		}
		if dependencyInStatus != nil {
			return nil, fmt.Errorf("found more then one vmo dependencies for workerpool %s in the worker provider status", workerPoolName)
		}
		dependencyInStatus = &dep
	}
```

pkg/controller/worker/machineset_vmo.go
```diff
@@ -172,6 +169,75 @@ func cleanupOrphanVMODependencies(ctx context.Context, client azureclient.Vmss,
...
+	// Third: No vmo for the worker pool was found on Azure. Need to create it.
+	faultDomainCount, err := azureapihelper.FindDomainCountByRegion(w.cloudProfileConfig.CountFaultDomains, w.worker.Spec.Region)
+	if err != nil {
+		return nil, err
+	}
+
+	newDependency, err := generateAndCreateVmo(ctx, vmoClient, workerPoolName, infrastructureStatus.ResourceGroup.Name, w.worker.Spec.Region, faultDomainCount)
+	if err != nil {
+		return nil, err
+	}
```
kon-angelo, line 233: Do we actually need to attempt to create the VMO here ? Wouldn't be enough if we just returned an error and waited for the next worker reconciliation instead of duplicating the creation logic ?

pkg/controller/worker/machineset_vmo.go
dkistner: Exactly, I need to capture the value. I need  define an explicit variable here otherwise `dependencyInStatus` would always point to the current `dep` value. I agree the name `runVar` is not optimal, but actually I don't need to do it for loop cycle in the beginning.

pkg/controller/worker/machineset_vmo.go
dkistner: If the Shoot is already in deletion phase then no next reconciliation flow will ever happen.
This actually also just some safety logic to ensure that a vmo exists + is part of the worker status when the machineclasses are generated.

pkg/controller/worker/machineset_vmo.go
kon-angelo: > but actually don't need to do it for loop cycle in the beginning

Yes, the code is pretty much optimal as it was. My suggestion was only to prevent this common pitfall in a future contribution. It is just one of those "easy-to-miss" things.

kon-angelo (approved):
/lgtm

pkg/controller/worker/machineset_vmo.go
dkistner: How about renaming the `runVar` in `depSnapshot` or something like this to make it easier to understand why we need this step?

kon-angelo: I approved anyway because this is clearly a matter of preference, but I like this suggestion a lot.

dkistner: Let's have this small improvement before we go on :)

dkistner: Done.

PR #278, opened 2021-04-09
depend on publish-trait, rather on injected step(s)

ccwienk:
prepare for switch to kaniko-builder


ialidzhikov (approved):
/lgtm

PR #279, opened 2021-04-09
Refactor pipeline def rel v1.18

ccwienk:
prepare switch to kaniko-builder

### DIFF
--- a/.ci/pipeline_definitions
+++ b/.ci/pipeline_definitions
@@ -35,7 +35,7 @@ gardener-extension-provider-azure:
         test-integration:
           execute:
           - test-integration.sh
-          depends:
+          trait_depends:
           - publish
           image: 'eu.gcr.io/gardener-project/gardener/testmachinery/testmachinery-run:stable'
       traits:


AndreasBurger:
This one targets the wrong branch, doesn't it?

ccwienk:
@AndreasBurger : correct - fixed this..

ccwienk:
I do not think the build-error is related to this diff

AndreasBurger:
The previous run was running against the base-branch on Concourse. The latest _did_ run against the correct branch, but still claims that `go.sum` was modified and a revendor needs to be run.

@ialidzhikov Can we ignore this? I don't see any change in the diff

ialidzhikov:
> @ialidzhikov Can we ignore this? I don't see any change in the diff

The verify step fails because the pipeline definition is not adapted accordingly after the go1.16 switch (cc @stoyanr @rfranzke ). For release-v1.18 it should use go1.15 but currently it uses go1.16 which causes the verify step to fail.
Do we really need this change cherry-picked into the release-v1.18 branch?

ccwienk:
> Do we really need this change cherry-picked into the release-v1.18 branch?
@ialidzhikov : yes, we do. actually, another option might be to explicitly configure the oci-builder to `concourse`. If none of those options is done, pipeline replication will break after defining `kaniko` as default builder.

How I see it, this branch is currently _broken_ (unrelated to my proposed PR). If we leave it like this, without an additional fix, it will _not_ be possible to create new versions from it.

Depending on whether or not this is an issue, I suggest to either:

- remove the branch (in case we do not plan to create a hotfix-release anyhow)
- merge this PR and ignore the fact the branch's state is actually broken (at least the infrastructure will remain ready for a potential hotfix-release)
- fix the build (+ accept this PR)
- remove the `.ci/pipeline_definitions` file (will remove the pipeline) - in this case, obviously discard this PR

I am fine with either option. However, I would personally prefer an option where we do not keep a broken branch.

ialidzhikov:
@ccwienk , I don't see any cherry-picks or cherry-pick candidates for the `release-v1.18` branch. That's why I assume it will be "garbage collected" relatively soon.
Meanwhile there is new `release-v1.19` branch and you can rebase this PR to be against `release-v1.19`.

rfranzke:
I deleted the `release-v1.18` branch with the rationale outlined by @ialidzhikov.

PR #284, opened 2021-04-14
Refactor pipeline def v1.19

ccwienk:
prepare switch to kaniko as default-builder


ialidzhikov (approved):
/lgtm

PR #292, opened 2021-05-07
Configure TCP egress for shoots

kon-angelo:

/area control-plane
/kind bug
/platform azure

When using the reverse cluster vpn model, there is no vpn-shoot service anymore created in the shoot. As a result the Azure LB is not configured to allow egress for TCP traffic when **not** using any NAT Gateway. As a result the cluster can never be created (or never stay healthy), as the kubelet cannot reach the API server. 

This PR creates a dummy TCP LB service, similar to the allow-udp-egress, to circumvent the situation.

Fixes #

**Release note**:

```bugfix operator
A new service `allow-tcp-egress` is created in the shoot cluster to configure TCP egress traffic when using the `reversed cluster vpn` feature.
```

### DIFF
--- a/charts/internal/shoot-system-components/charts/allow-egress/templates/allow-tcp-egress.yaml
+++ b/charts/internal/shoot-system-components/charts/allow-egress/templates/allow-tcp-egress.yaml
@@ -1,0 +1,25 @@
+# this is a technical service created to mitigate an issue with the TCP egress traffic for Shoots using
+# Azure Standard LoadBalancers (see https://github.com/gardener/gardener-extension-provider-azure/issues/1)
+apiVersion: v1
+kind: Service
+metadata:
+  name: allow-tcp-egress
+  namespace: kube-system
+  annotations:
+    azure.remedy.gardener.cloud/ignore: "true"
+    gardener.cloud/description: |
+      This is a technical Service created to mitigate an issue with the TCP egress traffic for Shoots using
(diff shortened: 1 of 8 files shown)


charts/internal/shoot-system-components/charts/allow-egress/templates/allow-tcp-egress.yaml
```diff
@@ -0,0 +1,26 @@
...
+metadata:
+  name: allow-tcp-egress
+  namespace: kube-system
+  annotations:
+    azure.remedy.gardener.cloud/ignore: "true"
+    gardener.cloud/description: |
+      This is a technical Service created to mitigate an issue with the TCP egress traffic for Shoots using
+      Azure Standard LoadBalancers (see https://github.com/gardener/gardener-extension-provider-azure/issues/1).
+      It is needed because the Standard LoadBalancers will block all outgoing TCP traffic if there is not at least
+      one open TCP Port. This is a temporary workaround and will be removed in a future version, once we have a
```
dkistner, line 14: Can we remove the statement for temporary workaround? I mean the NatGateway will stay optional, that means this service would be always required if there is no NatGateway used.

dkistner (commented):
Looks general good to me.
Just a minor comment.

As discussed off line a nice addition from resource consumption/cost point of view would be to deploy those allow services only if no NatGateway is configured. But let's handle that in a subsequent pr. Can we open a ticket for that?

dkistner (approved):
/lgtm

PR #297, opened 2021-05-12
Prepare PATH in prepare_release in advance

AndreasBurger:
The golang version installed by the vendored-in script is not found in path causing it to fall back to an older version.
As a fix/workaround, set PATH version in advance for now

### DIFF
--- a/.ci/prepare_release
+++ b/.ci/prepare_release
@@ -1,3 +1,6 @@
 #!/usr/bin/env bash
 
+# update PATH in preparation for golang download by the prepare_release script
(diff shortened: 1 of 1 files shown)


ialidzhikov (approved):
/lgtm

kon-angelo (approved):
Thank you!

rfranzke:
Is this also relevant for other extensions?

ialidzhikov:
> Is this also relevant for other extensions?

Yeah, I think so. So this PR is kind of a hack for provider-azure to unblock the failing release. I think we should have a proper fix in g/g where the script is maintained.

PR #308, opened 2021-05-19
Do not ensure apiserver flag `external-hostname`

dkistner:

/kind cleanup
/platform azure

Do not ensure the apiserver flag `--external-hostname` anymore.

Ref: https://github.com/gardener/gardener/pull/4067#issuecomment-844018152

/invite @ialidzhikov @AndreasBurger 

**Release note**:

```NONE

```

### DIFF
--- a/pkg/webhook/controlplaneexposure/ensurer.go
+++ b/pkg/webhook/controlplaneexposure/ensurer.go
@@ -94,7 +94,6 @@ func (e *ensurer) EnsureKubeAPIServerDeployment(ctx context.Context, gctx gconte
 
 	if c := extensionswebhook.ContainerWithName(new.Spec.Template.Spec.Containers, "kube-apiserver"); c != nil {
 		c.Command = extensionswebhook.EnsureStringWithPrefix(c.Command, "--advertise-address=", address)
-		c.Command = extensionswebhook.EnsureStringWithPrefix(c.Command, "--external-hostname=", address)
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


dkistner:
/unassign

@ialidzhikov PTAL :)

vpnachev:
What about `--advertise-address=`? It also cannot not be set by this extension when `APIServerSNI` gardenlet feature gate is enabled. I do not insist to remove it, but rather to improve the line above to not return error when the `kube-apiserver` service is not of type `LoadBalancer`.

AndreasBurger:
Tried to find out why we set the `external-hostname` flag back then, but unfortunately the relevant histories are not preserved. I assume it was necessary for the setup back then.
No objections to this change from my side if it isn't needed anymore.

ialidzhikov (approved):
/lgtm

dkistner:
> What about --advertise-address=? It also cannot not be set by this extension when APIServerSNI gardenlet feature gate is enabled. I do not insist to remove it, but rather to improve the line above to not return error when the kube-apiserver service is not of type LoadBalancer.

Yes we could improve here, may be with something like that:
```suggestion
	var apiServerService = &corev1.Service{ObjectMeta: metav1.ObjectMeta{Namespace: new.Namespace, Name: v1beta1constants.DeploymentNameKubeAPIServer}}
	address, err := kutil.GetLoadBalancerIngress(ctx, e.client, apiServerService)
	if err != nil {
		if apiServerService.Spec.Type == "" || apiServerService.Spec.Type == corev1.ServiceTypeLoadBalancer {
			return errors.Wrap(err, "could not get kube-apiserver service load balancer address")
		}
		return nil
	}
	if apiServerService.Spec.Type != corev1.ServiceTypeLoadBalancer {
		return nil
	}
```

But wouldn't it better to improve the `GetLoadBalancerIngress` function to return `*string, error` and to return `nil, nil` when service type is not `LoadBalancer`?

dkistner:
ahh good catch. No then we don't need it at all. Thank you.

vpnachev (approved):
That explains why I have not seen any error logs. I missed that check. 

/lgtm

rfranzke:
@ialidzhikov @vpnachev I'm not sure about the current state of this PR - is it ready to merge or are changes needed?

PR #321, opened 2021-06-09
Enhance validation of cloud provider secrets

stoyanr:

/area usability
/area ops-productivity
/kind enhancement
/platform azure

Enhances the validation of Azure cloud provider secrets:
* `subscriptionID`, `tenantID`, and `clientID` must be valid UUIDs (and therefore also can't contain leading or trailing whitespaces).
* `clientSecret` must not contain leading or trailing new lines, as they are known to cause issues. Other whitespace characters such as spaces are intentionally not checked for, since there I couldn't find any documentation indicating that they would not be valid.

Fixes #

**Release note**:

```other operator
Validation of Azure cloud provider secrets is enhanced to reject `subscriptionID`, `tenantID`, and `clientID` that are not valid UUIDs, and `clientSecret` that contain leading or trailing new lines.
```

### DIFF
--- a/pkg/apis/azure/validation/secrets.go
+++ b/pkg/apis/azure/validation/secrets.go
@@ -16,9 +16,16 @@ package validation
 
 import (
 	"fmt"
+	"regexp"
+	"strings"
 
 	"github.com/gardener/gardener-extension-provider-azure/pkg/azure"
+
 	corev1 "k8s.io/api/core/v1"
+)
+
+var (
+	guidRegex = regexp.MustCompile("^[0-9A-Fa-f]{8}-([0-9A-Fa-f]{4}-){3}[0-9A-Fa-f]{12}$")
 )
 
 // ValidateCloudProviderSecret checks whether the given secret contains a valid Azure client credentials.
@@ -32,6 +39,21 @@ func ValidateCloudProviderSecret(secret *corev1.Secret) error {
 		}
 		if len(val) == 0 {
 			return fmt.Errorf("field %q in secret %s cannot be empty", key, secretKey)
+		}
+		switch key {
+		case azure.SubscriptionIDKey, azure.TenantIDKey, azure.ClientIDKey:
+			// subscriptionID, tenantID, and clientID must be valid GUIDs,
+			// see https://docs.microsoft.com/en-us/rest/api/securitycenter/locations/get
+			if !guidRegex.Match(val) {
+				return fmt.Errorf("field %q in secret %s must be a valid GUID", key, secretKey)
+			}
+		case azure.ClientSecretKey:
+			// clientSecret must not contain leading or trailing new lines, as they are known to cause issues
+			// Other whitespace characters such as spaces are intentionally not checked for,
+			// since there is no documentation indicating that they would not be valid
+			if strings.Trim(string(val), "\n\r") != string(val) {
+				return fmt.Errorf("field %q in secret %s must not contain leading or traling new lines", key, secretKey)
+			}
 		}
 	}
 
--- a/pkg/apis/azure/validation/secrets_test.go
+++ b/pkg/apis/azure/validation/secrets_test.go
@@ -25,6 +25,13 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
+const (
+	subscriptionID = "b7ad693a-028a-422c-b064-d76c4586f2b3"
+	tenantID       = "ee16e592-3035-41b9-a217-958f8f75b740"
+	clientID       = "7fc4685d-3c33-40e6-b6bf-7857cab04300"
+	clientSecret   = "clientSecret"
+)
+
 var _ = Describe("Secret validation", func() {
 
 	DescribeTable("#ValidateCloudProviderSecret",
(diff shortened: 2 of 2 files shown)


pkg/apis/azure/validation/secrets.go
```diff
@@ -33,6 +36,20 @@ func ValidateCloudProviderSecret(secret *corev1.Secret) error {
 		if len(val) == 0 {
 			return fmt.Errorf("field %q in secret %s cannot be empty", key, secretKey)
 		}
+		switch key {
+		case azure.SubscriptionIDKey, azure.TenantIDKey, azure.ClientIDKey:
+			// subscriptionID, tenantID, and clientID must be valid UUIDs
```
vpnachev, line 41: Could you point the source of this information?

ialidzhikov: At least for subsciption ID, I found that several Azure REST APIs reveal the following regex - `^[0-9A-Fa-f]{8}-([0-9A-Fa-f]{4}-){3}[0-9A-Fa-f]{12}$`. See https://docs.microsoft.com/en-us/rest/api/securitycenter/locations/get, https://docs.microsoft.com/en-us/rest/api/securitycenter/tasks/get-resource-group-level-task and few others Azure REST API docs.
I assume that this regex fits with the validation performed by `github.com/google/uuid`.

pkg/apis/azure/validation/secrets_test.go
```diff
@@ -39,86 +46,126 @@ var _ = Describe("Secret validation", func() {
...
-				azure.ClientIDKey:       []byte("cliendID"),
+				azure.SubscriptionIDKey: []byte(subscriptionID),
+				azure.TenantIDKey:       []byte(tenantID),
+				azure.ClientIDKey:       []byte(clientID),
 				azure.ClientSecretKey:   {},
 			},
 			HaveOccurred(),
 		),
 
+		Entry("should return error when the client secret contains a trailing new line",
```
ialidzhikov, line 153: Nit: for completeness you could add an entry that tests client secret containing a leading new line.

pkg/apis/azure/validation/secrets.go
stoyanr: I could not find authoritative documentation (from Microsoft) stating this explicitly. What I could find are various documents stating that these IDs come from "Azure Active Directory", as well as references in documents (both from Microsoft and 3rd parties) referring to them as "GUIDs", see e.g. 
* https://docs.microsoft.com/en-us/powershell/module/azurerm.resources/get-azurermadapplication?view=azurermps-6.13.0
* https://stackoverflow.com/questions/56551171/is-azure-aad-application-id-unique-in-whole-azure
* https://docs.bmc.com/docs/cloudlifecyclemanagement/46/setting-up-a-tenant-id-client-id-and-client-secret-for-azure-resource-manager-provisioning-669202145.html#:~:text=The%20subscription%20ID%20is%20a,along%20with%20the%20subscription%20ID.

We also know from looking at actual credentials that they are always GUIDs.

I know this is a bit weak. If we want to be on the safe side, we should perhaps refrain from validating as GUIDs and only check for leading / trailing whitespace. However, if we validate them as GUIDs we will also be able to prevent copying / pasting a GUID and making a mistake. In practice, even if the current GUIDs turns out to be an implementation detail, it's one that's not very likely to change. So I would vote for keeping the GUID check.

@rfranzke Any comments?

pkg/apis/azure/validation/secrets_test.go
stoyanr: I don't see much value in this, it's basically checking that `strings.Trim` behaves to specification. Then I would also have to check for `\r` etc. I would leave it out unless you insist.

pkg/apis/azure/validation/secrets.go
stoyanr: @ialidzhikov Thanks! I added this reference to the code. I also reworked the validation to use a regex match for the above pattern, instead of `github.com/google/uuid` since I noticed that it allows uuids with formats different from the above that would likely fail on Azure (I checked one of them and it failed).

rfranzke (approved):
/lgtm
/squash
/invite @dkistner @kon-angelo

kon-angelo (approved):
/lgtm 

extra points for the clear documentation behind the rules :)

pkg/apis/azure/validation/secrets.go
ialidzhikov: > @ialidzhikov Thanks! I added this reference to the code. I also reworked the validation to use a regex match for the above pattern, instead of github.com/google/uuid since I noticed that it allows uuids with formats different from the above that would likely fail on Azure (I checked one of them and it failed).

Above I only shared the regex for the subsciption ID. Does the same regex apply for tenant ID and client ID?

kon-angelo: I feel a bit conflicted now. I checked our landscapes to see examples for `clientID` and `tenantID` and although they match the regex here - so no immediate issues from merging the PR - @ialidzhikov has a point we don't have official documentation for these regexes `cID` and `tID`. Not one that I could find at least.

kon-angelo: I discussed this with our Azure folks and since all 3 fields are GUIDs, the same restrictions apply equally. So the PR should be fine as it is.

PR #327, opened 2021-06-26
Update kubernetes-sigs/azuredisk-csi-driver

ialidzhikov:
**Release note**:

```other user
The following image is updated (see [CHANGELOG](https://github.com/kubernetes-sigs/azuredisk-csi-driver/releases/tag/v1.2.0) for more details):
- mcr.microsoft.com/k8s/csi/azuredisk-csi: v1.1.1 -> v1.2.0
```

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -44,7 +44,7 @@ images:
 - name: csi-driver-disk
   sourceRepository: github.com/kubernetes-sigs/azuredisk-csi-driver
   repository: mcr.microsoft.com/k8s/csi/azuredisk-csi
-  tag: "v1.1.1"
+  tag: "v1.2.0"
 - name: csi-driver-file
   sourceRepository: github.com/kubernetes-sigs/azurefile-csi-driver
   repository: mcr.microsoft.com/k8s/csi/azurefile-csi


charts/images.yaml
```diff
@@ -44,7 +44,7 @@ images:
 - name: csi-driver-disk
   sourceRepository: github.com/kubernetes-sigs/azuredisk-csi-driver
   repository: mcr.microsoft.com/k8s/csi/azuredisk-csi
-  tag: "v1.1.1"
+  tag: "v1.2.0"
```
vpnachev, line 47: Do you want drop support for k8s <= 1.16?
I might have wrong understanding of the [this](https://github.com/kubernetes-sigs/azuredisk-csi-driver#container-images--kubernetes-compatibility).
On the other side, there is versions 1.3.0 and 1.4.0.
Could you add short explanation why you have chosen to update to version 1.2.0?

ialidzhikov: > Do you want drop support for k8s <= 1.16?
> I might have wrong understanding of the [this](https://github.com/kubernetes-sigs/azuredisk-csi-driver#container-images--kubernetes-compatibility).

On azure, azuredisk-csi or in general CSI is enabled only for v1.21 clusters. So these version constraints don't affect us currently.

> On the other side, there is versions 1.3.0 and 1.4.0.
> Could you add short explanation why you have chosen to update to version 1.2.0?

Initially user reported that he/she tried to use `Premium_ZRS` disks on Kubernetes v1.21. The operation failed with reason: 

```
  "error": {
    "code": "InvalidParameter",
    "message": "Storage account type 'Premium_ZRS' is supported by Microsoft.Compute API version 2020-12-01 and above.",
    "target": "dataDisk.managedDisk.storageAccountType"
  }
```

Then @MSSedusch and @kon-angelo found that support for ZRS disks is added in azuredisk-csi@v1.2.0 - see https://github.com/kubernetes-sigs/azuredisk-csi-driver/releases/tag/v1.2.0. Of course, I known that there are even newer versions but currently my focus was first to update to a version with ZRS support to tackle the issue I mentioned.

vpnachev: Thanks for the info!

vpnachev (approved):
/lgtm

MSSedusch:
@ialidzhikov I would prefer to update it to CSI 1.4.0. @dkistner was planning to do that anyway since we want to test the parallel disk attach/detach which was added in 1.3.0

ialidzhikov:
> @ialidzhikov I would prefer to update it to CSI 1.4.0. @dkistner was planning to do that anyway since we want to test the parallel disk attach/detach which was added in 1.3.0

I now tried to play with azuredisk-csi-driver@v1.4.0. There is a quite confusing change (at least for me) - https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/831. It basically introduces a new flag `--disable-avset-nodes` (defaults to true). I played with azuredisk-csi-driver@v1.4.0 and for VMO clusters (that don't have availability set) the driver is failing with:

```
I0628 11:12:58.923046       1 azure_vmss.go:343] Can not extract scale set name from providerID (/subscriptions/<id>/resourceGroups/shoot--foo--azure-local/providers/Microsoft.Compute/virtualMachines/shoot--foo--azure-local-cpu-worker-57948-nl9mk), assuming it is managed by availability set: not a vmss instance
```

(The log is info one, but attach call to the driver is failing with similar `not a vmss instance` error)

Probably this can be fixed by passing `--disable-avset-nodes=false` for a VMO cluster but I don't think this is the right approach. As far a I could dig into, the issue is coming from https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/703d60d110b16075900eb1ec10e394c20289cce9/vendor/sigs.k8s.io/cloud-provider-azure/pkg/provider/azure_vmss.go#L47. It is expecting `.../Microsoft.Compute/virtualMachineScaleSets/...` while the instance on my side is `.../Microsoft.Compute/virtualMachines/...`.


In general, I don't see anything wrong on upgrading first to `azuredisk-csi-driver@v1.2.0` and then upgrading to azuredisk-csi-driver@v1.4.0 when we clarify the issue with VMO + `--disable-avset-nodes=true`

/cc @dkistner

dkistner:
The machine belonging to a vmo (aka vmss with flexible orchestration mode) are regular machines and managed independently of the vmss. So those machines are managed on Azure as `Microsoft.Compute/virtualMachines`.

In contrast, the machines of vmss with uniform orchestration mode are managed on Azure as `Microsoft.Compute/virtualMachineScaleSets`. 

Therefore I think it is ok to set `--disable-avset-nodes=false` for vmo clusters as their machines are regular machines.

@MSSedusch Do you have a recommendation here?

dkistner (approved):
/lgtm

We will update to `v1.4.0` in a subsequent pr. cc @MSSedusch

MSSedusch:
@dkistner I talked to @HappyTobi yesterday and we will look into the issue with CSI 1.4.0 and VMO. I don't think setting disable-avset-nodes is a good idea. We had a quick look at the coding and in some cases, setting this property makes the driver assume it is a VMSS cluster which is not the case.

PR #331, opened 2021-06-30
Zone redundant NAT Gateways

kon-angelo:

/area control-plane
/kind enhancement
/platform azure

This PR allows the deployment of shoots with a new network setup for Azure. The new setup is akin to other extensions where it is possible to have finer-grained control over the networking of each availability zone. The main motivation is to have zone redundant NAT Gateways. Prior to this PR we used a single subnet and thus a single NAT Gateway instance. Using the new setup each subnet can potentially have its own NAT gateway instance and separate public IPs.

In addition, there is a migration path from our current zoned setup to this "enhanced" zone setup. For this to be possible, we impose the constraint that the first zone specified in the infrastructure must match the CIDR range of the existing `workers` subnet. The main reason is to preserve the subnet of an existing cluster and hence prevent having to teardown all the existing VMs to recreate the network setup. Instead, by preserving the original subnet we can have a controlled node migration (via MCM) to the new `machineClasses` and ensure minimal downtime during the migration.

Fixes #https://github.com/orgs/gardener/projects/7#card-49971123

Currently missing:

+ [x] documentation
+ [x] a new terraform `azurerm` provider. 

Unfortunately there is a major bug with with TF provider for Azure that we encountered ([link](https://github.com/terraform-providers/terraform-provider-azurerm/issues/2489)) -  when creating multiple "association" resources, the terraform provider is permanently stuck due to locking issues. With the current (at the time of writing) version we use it is happening consistently. The "custom" version of the terraformer used in this PR is build based on a newer version (2.60) and while the occurrence rate has fallen, it is still around 50% according to my tests and I consider it a showstopper until fixed. The custom terraform image is there to facilitate reviewers and testers. 

We have a PR with a potential fix waiting for review (https://github.com/terraform-providers/terraform-provider-azurerm/pull/12267).

**Release note**:

```other operator
Azure provider now supports a new network setup that allows for zone redundant NAT Gateways.
```


pkg/controller/worker/machines.go
```diff
@@ -238,10 +225,14 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 
 		// VMO
 		if vmoDependency != nil {
+			workerPoolHash, err := w.generateWorkerPoolHash(pool, infrastructureStatus, vmoDependency, nil)
```
dkistner, line 228: The `workerPoolHash` is generated in 4 of 5 cases with same inputs. How about generating it once and override it in the `subnet != nil` case? That would improve in regards to DRY and eliminate at least one else case below.

Or even better in the `generateMachineClassAndDeployment()` function.

pkg/controller/worker/machines.go
```diff
@@ -270,11 +266,40 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
...
+				workerPoolHash string
+			)
+
+			if infrastructureStatus.Networks.Topology == azureapi.TopologyZonal {
+				subnetIndex, nodesSubnet, err := azureapihelper.FindSubnetByPurposeAndZone(infrastructureStatus.Networks.Subnets, azureapi.PurposeNodes, zone)
+				if err != nil {
+					return err
+				}
+				subnetName = nodesSubnet.Name
+				if subnetIndex == 0 {
```
dkistner, line 280: This is just required as we keep during network layout migration the machines assigned to existing subnet, right?

dkistner (commented):
I played a bit around with it, but not yet in detail with the migration part.

After my first detailed scan I have a few suggestions regarding the api of the `InfrastructureConfig` and some additional comments below.

1. Is the `.zoned=true` field really required when setting `.networks.zones[] > 0`? I know we need the `.zoned` field in the `InfrastructureStatus` as it is input for subsequent controllers, but we could also calculate the field in the status respectively.
2. Do we need to set the `zone` of a natgateway in `.networks.zones[].natGateway.zone`? I mean the zone is implicitly set by `.networks.zones[].name`. Same is probably true for `.networks.zones[].natGateway.ipAddresses[].zone`. I know removing that would mean to introduce additional strucs but it would be cleaner from an api perspective in my opinion. WDYT?

kon-angelo:
> 1. Is the .zoned=true field really required when setting .networks.zones[] > 0? I know we need the .zoned field in the InfrastructureStatus as it is input for subsequent controllers, but we could also calculate the field in the status respectively.
> 2. Do we need to set the zone of a natgateway in .networks.zones[].natGateway.zone? I mean the zone is implicitly set by .networks.zones[].name. Same is probably true for .networks.zones[].natGateway.ipAddresses[].zone. I know removing that would mean to introduce additional strucs but it would be cleaner from an api perspective in my opinion. WDYT?

1. we can infer the `zoned`  if the `network.zones` exit, however having zones specified and `zoned` defaulting to false would be misleading, so we would have to change `zoned` to a `*bool`. It is used in a lot of places in the code and we would nil checks etc., which is why I opted to validate the configuration instead.

2. The setting of the zone was done in spirit of the current relationship NatGateways have with PublicIPs. PIPs can only be set on NATGWs that specify a zone and the PIPs also have to specify a zone (instead of inferring it from the NAT). You are correct that in the `zones` case we don't need them necessarily. Let's go with your suggestion.

@dkistner

pkg/apis/azure/validation/infrastructure.go
```diff
@@ -29,6 +31,44 @@ const (
...
+func validateInfrastructureConfigZones(oldInfra, infra *apisazure.InfrastructureConfig, zones []gardencorev1beta1.AvailabilityZone, fld *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	azureZones := sets.NewString()
+	for _, zone := range zones {
+		azureZones.Insert(zone.Name)
+	}
+
+	for i, zone := range infra.Networks.Zones {
+		if oldInfra != nil && len(oldInfra.Networks.Zones) > i && oldInfra.Networks.Zones[i].Name == zone.Name {
```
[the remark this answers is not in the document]
kon-angelo: Yes, this is to prevent the case where removing the zone from the `CloudProfile` would render the shoot unable to update. If the zone was already part of the previous shoot configuration, we continue without checking if it still exists in the CP.

pkg/controller/worker/machines.go
kon-angelo: Yes. So the subnetName was not used previously when calculating the workerpool hash.
This change requires that:
a) machines in zone["0"] (the one keeping the original subnet) are correctly placed and thus don't need any change in their hash (avoid triggering rolling updates).
b) all other machines that reside in zones but may have been using the previously single subnet to be rolled to use the new subnets.

The index here is used to make the distinction on whether to use or not the subnet as part of the hash.

pkg/internal/terraform.go
```diff
@@ -89,6 +89,7 @@ func NewTerraformer(logger logr.Logger, restConfig *rest.Config, purpose string,
 		SetTerminationGracePeriodSeconds(630).
 		SetDeadlineCleaning(5 * time.Minute).
 		SetDeadlinePod(15 * time.Minute).
+		SetLogLevel("DEBUG").
```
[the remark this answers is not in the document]
kon-angelo: Leftover for debugging the deadlock issue. Thanks for the catch.

pkg/apis/azure/validation/infrastructure.go
```diff
@@ -232,3 +428,7 @@ func isDefaultVnetConfig(vnetConfig *apisazure.VNet) bool {
 	}
 	return false
 }
+
+func isUsingInfrastructureZones(netConfig *apisazure.NetworkConfig) bool {
```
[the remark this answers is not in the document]
kon-angelo: Much better thank you!

pkg/apis/azure/v1alpha1/types_infrastructure.go
```diff
@@ -48,13 +48,16 @@ type NetworkConfig struct {
 	// VNet indicates whether to use an existing VNet or create a new one.
 	VNet VNet `json:"vnet"`
 	// Workers is the worker subnet range to create (used for the VMs).
-	Workers string `json:"workers"`
+	// +optional
+	Workers *string `json:"workers"`
```
ialidzhikov, line 52: As this PR makes the field optional, don't we need to have `,omitempty`?

From https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md:

> Optional fields should have the `,omitempty` json tag; fields are interpreted as being required otherwise.

ialidzhikov (commented):
I cannot provide much valuable feedback for the new networking deployment model and its migration. I am leaving few comments of general purpose type.

pkg/internal/infrastructure/terraform.go
```diff
@@ -167,16 +172,6 @@ func ComputeTerraformerTemplateValues(
...
 
-	natGatewayConfig, createNatGateway := generateNatGatewayValues(config)
-
-	// Checks if the Gardener managed NatGateway public ip needs to be migrated.
-	// TODO(natipmigration) This can be removed in future versions when the ip migration has been completed.
-	natGatewayIPMigrationRequired, err := isNatGatewayIPMigrationRequired(infra, config)
-	if err != nil {
-		return nil, err
-	}
-	natGatewayConfig["migrateNatGatewayToIPAssociation"] = natGatewayIPMigrationRequired
```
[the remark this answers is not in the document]
kon-angelo: I restored the fields. For the `migrateNatGatewayToIPAssociation` I  already checked the landscapes and did not find anyplace where they used Nat and it was set to true so I am confident that we won't have any issue for our landscapes. But you are correct and let's do properly in a separate PR.

bortk:
Hi,
What is the status of this PR? We are waiting for it to implement multi-AZ on Azure.
BR,
Boris

dkistner (commented):
First part of my detailed scan.
I still need to check the validation part.

pkg/apis/azure/validation/infrastructure.go
```diff
@@ -17,18 +16,72 @@ package validation
...
 const (
 	natGatewayMinTimeoutInMinutes int32 = 4
 	natGatewayMaxTimeoutInMinutes int32 = 120
 )
 
+// ValidateInfrastructureConfigAgainstCloudProfile validates the InfrastructureConfig against the CloudProfile.
+func ValidateInfrastructureConfigAgainstCloudProfile(oldInfra, infra *apisazure.InfrastructureConfig, shootRegion string, cloudProfile *gardencorev1beta1.CloudProfile, fld *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if len(infra.Networks.Zones) == 0 {
```
dkistner, line 39: minor hint: You could use your helper `IsUsingSingleSubnetLayout()` here ;)

pkg/apis/azure/validation/infrastructure.go
```diff
@@ -62,28 +115,159 @@ func ValidateInfrastructureConfig(infra *apisazure.InfrastructureConfig, nodesCI
...
+			allErrs = append(allErrs, field.Forbidden(workersPath, "natGateway cannot be specified when workers field is missing"))
+		}
+		if len(config.ServiceEndpoints) > 0 {
+			allErrs = append(allErrs, field.Forbidden(workersPath, "serviceEndpoints cannot be specified when workers field is missing"))
+		}
+
+		allErrs = append(allErrs, validateZones(config.Zones, nodes, pods, services, zonesPath)...)
 	}
 
+	allErrs = append(allErrs, validateVnetConfig(&config, infra.ResourceGroup, workerCIDR, nodes, pods, services, zonesPath, vNetPath)...)
```
dkistner, line 183: Can you move this before the `if else` block?
Then we could eliminate the `if else` by returning after the `helper.IsUsingSingleSubnetLayout` if block is completed.

dkistner (commented):
Now also done the validation part.

After an intensive pair review on that on functional and code level I have only minor comments lefts. After checking and addressing them where necessary I think we are finally good with this PR.

pkg/webhook/infrastructure/mutate.go
```diff
@@ -0,0 +1,137 @@
...
+
+	// if the old configuration is not zoned and it is already using multiple-subnet layout, no mutation is necessary.
+	if !oldProviderCfg.Zoned || len(oldProviderCfg.Networks.Zones) > 0 {
+		return nil
+	}
+
+	// take care of clusters that have not been reconciliated for a long time (hibernated etc). In this case they may
+	// not have the Layout field populated, so instead rely on the old spec. If the cluster was not a zoned one skip it.
+	if oldProviderStatus != nil &&
+		oldProviderStatus.Networks.Layout != "" &&
```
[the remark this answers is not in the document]
kon-angelo: Makes sense, I had the check arranged in ta different manner but I forgot to update the comment. 

For your question, what the code is doing here, is checking "if the `layout` has a value and if it is not the expected one for the mutation" -> terminate early. Otherwise continue with the mutation.

dkistner (approved):

PR #332, opened 2021-07-02
Fix dynamic provisioning with the etcd StorageClass on Kubernetes v1.21

ialidzhikov:
/area storage
/kind bug

Currently on a Kubernetes v1.21 (CSI enabled) Seed, the provisioning of etcd volume (PVC with the `gardener.cloud-fast` StorageClass) fails with:

```
E0702 10:16:07.216628       1 volume_store.go:144] error saving volume pv-shoot--foo--bar-01956d9d-bdbf-4602-b1b6-6c1b20a86349: PersistentVolume "pv-shoot--foo--bar-01956d9d-bdbf-4602-b1b6-6c1b20a86349" is invalid: [spec.azureDisk.kind: Unsupported value: "managed": supported values: "Dedicated", "Managed", "Shared", spec.azureDisk.diskURI: Unsupported value: "/subscriptions/<omitted>/resourceGroups/shoot--foo--bar/providers/Microsoft.Compute/disks/pv-shoot--foo--bar-01956d9d-bdbf-4602-b1b6-6c1b20a86349": supported values: "https://{account-name}.blob.core.windows.net/{container-name}/{disk-name}.vhd"]
```

Basically this  `gardener.cloud-fast` StorageClass is still using the in-tree provisioner `kubernetes.io/azure-disk`. Currently the CSI translation library fails to translate `kind: managed`.
More details on https://github.com/kubernetes/kubernetes/issues/103433.

This PR removes `kind: managed` from the ETCD StorageClass. The default `kind` is already `managed` from Kubernetes v1.12 - ref https://github.com/kubernetes/kubernetes/pull/67483.

**Release note**:

```bugfix operator
An issue causing dynamic provisioning with the `gardener.cloud-fast` StorageClass to fail on Kubernetes v1.21 (or any other version with CSI enabled) is now fixed.
```


vpnachev (approved):
/lgtm

vpnachev:
Those that are still using `kind: managed` in https://github.com/gardener/gardener-extension-provider-azure/blob/master/charts/internal/shoot-storageclasses/templates/storageclasses.yaml are fine because the `provisioner: disk.csi.azure.com`, right?

ialidzhikov:
> Those that are still using `kind: managed` in https://github.com/gardener/gardener-extension-provider-azure/blob/master/charts/internal/shoot-storageclasses/templates/storageclasses.yaml are fine because of the `provisioner: disk.csi.azure.com`, right?

Right.

PR #352, opened 2021-07-30
Stop using `github.com/pkg/errors`

### LINKED ISSUE #337: Stop using github.com/pkg/errors

/kind enhancement
/priority 3
/platform azure

 Similar to gardener/gardener#4280 we should be using Go native error wrapping (available since GO 1.13).
```
grep -r '"github.com/pkg/errors"' | grep -v vendor/ | cut -f 1 -d ':' | cut -d '/' -f 1-3 | sort | uniq -c | sort
      1 pkg/apis/azure
      1 pkg/controller/infrastructure
      1 pkg/webhook/controlplane
      1 pkg/webhook/controlplaneexposure
      1 test/tm/generator.go
      2 pkg/controller/controlplane
      3 pkg/controller/worker
```

**Why is this needed**:
Getting rid of vendors in favor of using stdlib is always nice. Others seem to do this as well - kubernetes/kubernetes#103043 and containerd/console#54.

dkistner:

/kind enhancement
/platform azure

Similar to gardener/gardener#4280 we should be using Go native error wrapping (available since GO 1.13).

Fixes #337

**Release note**:

```NONE

```

/invite @vanjiii

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,7 +27,6 @@ require (
 	github.com/onsi/ginkgo v1.16.4
 	github.com/onsi/gomega v1.13.0
 	github.com/pierrec/lz4 v2.5.1+incompatible // indirect
-	github.com/pkg/errors v0.9.1
 	github.com/sirupsen/logrus v1.7.0
 	github.com/spf13/cobra v1.1.3
 	github.com/spf13/pflag v1.0.5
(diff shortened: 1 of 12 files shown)


kon-angelo (approved):
/lgtm

Just my 2cents but the `pkg/errors` is not completely obsolete as it has a slightly different behaviour than the native go's error wrapping: it records the stack trace - it can still be useful at places just not as the wrap mechanism.

Also in subsequent PRs we could attempt to clean some wrappings to use `%v` since the wrapped errors are not really useful to the callers e.g. the `tm/generator`.

vanjiii (approved):

vpnachev (approved):
/lgtm

PR #353, opened 2021-07-30
Remove hardcoded os disk type check in the worker controller.

### LINKED ISSUE #190: Clean up the handling for .spec.provider.workers[].machine.volume.type=standard
https://github.com/gardener-attic/gardener-extensions/pull/401 introduces the following code fragment for backwards-compatibility reasons:

https://github.com/gardener/gardener-extension-provider-azure/blob/b4c364899891e03886d607f2578409e256bbd9f4/pkg/controller/worker/machines.go#L306-L315

Prior to https://github.com/gardener-attic/gardener-extensions/pull/401, the azure machines were always with the default os disk belonging to the requested machine type.

We already have in place the validation that prevents creation of a new cluster with volume type != `["Standard_LRS", "StandardSSD_LRS", "Premium_LRS"]`.

```
$ k apply -f ~/bar.yaml --dry-run=server
Error from server (Forbidden): error when creating "/Users/foo/bar.yaml": shoots.core.gardener.cloud "test" is forbidden: [spec.provider.workers[0].volume.type: Unsupported value: core.Volume{Name:(*string)(nil), Type:(*string)(0xc013ff5820), VolumeSize:"50Gi", Encrypted:(*bool)(nil)}: supported values: "Standard_LRS", "StandardSSD_LRS", "Premium_LRS"]
```

We still have a small number of legacy Shoots which still use ` .spec.provider.workers[].machine.volume.type=standard`. We should actively ping them to migrate (a rolling update of the Nodes will be caused by change of `volume.type`).

After that we can clean up the above explicit check for `["Standard_LRS", "StandardSSD_LRS", "Premium_LRS"]`.

/kind cleanup
/platform azure
/priority normal
/cc @dkistner

dkistner:

/kind cleanup
/platform azure

Clean up a hardcoded os disk type check in the worker controller.

Fixes #190

I will keep this in draft until we have migrated all cluster to use a valid os disk type.
For all other who are using the `provider-azure` extension I added a `breaking` release note to make them aware of this change.

**Release note**:

```operator breaking
An explizit check for the os disk type in the worker controller has been removed. Please ensure that existing Azure Shoots which specify a type for the os disk use a valid one. The os disk type is configured per worker pool via `.spec.provider.workers[].volume.types`. The available volume types can be looked up in the `CloudProfile` via `spec.volumeTypes[]`. If the Shoot does not specify an os disk type then the default disk type for the corresponding machine type will be used. Please check the Azure machine documentation for more detail.
```

/invite @ialidzhikov @kon-angelo

### DIFF
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -418,14 +418,7 @@ func computeDisks(pool extensionsv1alpha1.WorkerPool, dataVolumesConfig []azurea
 	osDisk := map[string]interface{}{
 		"size": volumeSize,
 	}
-	// In the past the volume type information was not passed to the machineclass.
-	// In consequence the Machine controller manager has created machines always
-	// with the default volume type of the requested machine type. Existing clusters
-	// respectively their worker pools could have an invalid volume configuration
-	// which was not applied. To do not damage existing cluster we will set for
-	// now the volume type only if it's a valid Azure volume type.
-	// Otherwise, we will still use the default volume of the machine type.
-	if pool.Volume.Type != nil && (*pool.Volume.Type == "Standard_LRS" || *pool.Volume.Type == "StandardSSD_LRS" || *pool.Volume.Type == "Premium_LRS") {
+	if pool.Volume != nil && pool.Volume.Type != nil {
 		osDisk["type"] = *pool.Volume.Type
 	}
 


kon-angelo (approved):
/lgtm

kon-angelo (dismissed):
/lgtm

kon-angelo (dismissed):
/lgtm

hebelsan (approved):
/lgtm

kon-angelo (approved):
/lgtm

Issue #373, opened 2021-09-21
Consider switching to the out-of-tree Azure cloud-controller-manager


ialidzhikov:
**How to categorize this issue?**

/area control-plane
/kind enhancement
/priority 3
/platform azure

**What would you like to be added**:
Before https://github.com/gardener/gardener-extension-provider-azure/pull/72, we were already using the out-of-tree Azure cloud-controller-manager for a short period of time. With https://github.com/gardener/gardener-extension-provider-azure/pull/72, we switched to our fork of the in-tree Azure CCM as the out-of-tree one was still considered as alpha and not ready for production.

But things changed and out-of-tree CCM claims to be GA now:
> cloud-provider-azure has been GA since v1.0.0

**Why is this needed**:
To get rid of our CCM fork (https://github.com/gardener/cloud-provider-azure) and all the maintenance effort related to it.

ialidzhikov:
Fixed with https://github.com/gardener/gardener-extension-provider-azure/pull/431.

/close

Issue #374, opened 2021-09-21
Cannot delete last LoadBalancer service when using VMO


kon-angelo:
**How to categorize this issue?**

/area control-plane
/kind bug
/priority 3
/platform azure

**What happened**:

Deleting the last loadbalancer service will fail with error:

```
  Warning  SyncLoadBalancerFailed   3m27s (x11 over 6m35s)  service-controller               Error syncing load balancer: failed to delete load balancer: not a vmss instance
```

**What you expected to happen**:

For the service to be deleted

**How to reproduce it (as minimally and precisely as possible)**:

1. Create a VMO cluster
2. Create a new service of type loadbalancer
3. Try to delete the service.

Be mindful, that if the shoot is not using the reverse VPN tunneling, then the "last" service will be the _vpn-shoot_ service. In that case the skip step 2 and the_vpn-shoot_ service should be deleted 

**Anything else we need to know?**:

The error seems to not be reproducible with the latest `cloud-provider-azure` version (v1.1.1). Solving https://github.com/gardener/gardener-extension-provider-azure/issues/373 is a potential mitigation.  

**Environment**:

- Gardener version (if relevant):
- Extension version: v1.21.x
- Kubernetes version (use `kubectl version`): 1.21.x
- Cloud provider or hardware configuration:
- Others:

ialidzhikov:
Ref https://github.com/kubernetes/kubernetes/pull/105185 (credits to @kon-angelo)

PR #376, opened 2021-09-21
[UPDATE] use offical image from ms

HappyTobi:

/area robustness
/kind enhancement
/platform azure

Because the official "new" way how the implementation will be updated was moved from the "inner-tree" to the k8s-sigs repos, so it would be easier to fetch the implementation direct from there.

Fixes: https://github.com/gardener/gardener-extension-provider-azure/issues/373
Multiple issues of shoots that was deployed on vmo / vmss-flex.
(we see some issues in the latest 4-6 months that was already fixed at the k8s-sigs repo)

What kind of way to you prefer to run the binary of the container.
Just move the args to the "args" key of the container spec or changing / adding the new entrypoint.

**Release note**:

```other operator
category: feature
target_group: dependency
```

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -29,14 +29,14 @@ images:
   tag: "v1.20.10"
   targetVersion: "1.20.x"
 - name: cloud-controller-manager
-  sourceRepository: github.com/gardener/cloud-provider-azure
-  repository: eu.gcr.io/gardener-project/kubernetes/cloud-provider-azure
-  tag: "v1.21.4"
+  sourceRepository: https://github.com/kubernetes-sigs/cloud-provider-azure
+  repository: mcr.microsoft.com/oss/kubernetes/azure-cloud-controller-manager
+  tag: "v1.1.1"
   targetVersion: "1.21.x"
 - name: cloud-controller-manager
-  sourceRepository: github.com/gardener/cloud-provider-azure
-  repository: eu.gcr.io/gardener-project/kubernetes/cloud-provider-azure
-  tag: "v1.22.0"
+  sourceRepository: https://github.com/kubernetes-sigs/cloud-provider-azure
+  repository: mcr.microsoft.com/oss/kubernetes/azure-cloud-controller-manager
+  tag: "v1.1.1"
   targetVersion: ">= 1.22"
 - name: machine-controller-manager
   sourceRepository: github.com/gardener/machine-controller-manager
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -31,7 +31,7 @@ images:
 - name: cloud-controller-manager
   sourceRepository: https://github.com/kubernetes-sigs/cloud-provider-azure
   repository: mcr.microsoft.com/oss/kubernetes/azure-cloud-controller-manager
-  tag: "v1.1.1"
+  tag: "v1.0.5"
   targetVersion: "1.21.x"
 - name: cloud-controller-manager
   sourceRepository: https://github.com/kubernetes-sigs/cloud-provider-azure
--- a/charts/internal/seed-controlplane/charts/cloud-controller-manager/templates/cloud-controller-manager.yaml
+++ b/charts/internal/seed-controlplane/charts/cloud-controller-manager/templates/cloud-controller-manager.yaml
@@ -40,7 +40,7 @@ spec:
         {{- if semverCompare "< 1.17" .Values.kubernetesVersion }}
         - /hyperkube
         - cloud-controller-manager
-        {{- else if semverCompare ">=1.21" .Values.kubernetesVersion }}
+        {{- else if semverCompare ">= 1.21" .Values.kubernetesVersion }}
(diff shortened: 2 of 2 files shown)


gardener-robot-ci-3:
Thank you @HappyTobi for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

charts/images.yaml
```diff
@@ -29,14 +29,14 @@ images:
...
+  tag: "v1.1.1"
   targetVersion: "1.21.x"
 - name: cloud-controller-manager
-  sourceRepository: github.com/gardener/cloud-provider-azure
-  repository: eu.gcr.io/gardener-project/kubernetes/cloud-provider-azure
-  tag: "v1.22.0"
+  sourceRepository: https://github.com/kubernetes-sigs/cloud-provider-azure
+  repository: mcr.microsoft.com/oss/kubernetes/azure-cloud-controller-manager
+  tag: "v1.1.1"
   targetVersion: ">= 1.22"
```
rfranzke, line 40:
```suggestion
  sourceRepository: github.com/kubernetes-sigs/cloud-provider-azure
  repository: mcr.microsoft.com/oss/kubernetes/azure-cloud-controller-manager
  tag: "v1.1.1"
  targetVersion: ">= 1.21"
```

rfranzke (commented):
Thanks @HappyTobi!
Are we certain that `v1.1.1` is fully compatible with the `release-v1.21` and `release-v1.22` branches of the in-tree Azure provider?
Also, can you please provide a proper release note?

charts/images.yaml
kon-angelo: If I read the [matrix](https://github.com/kubernetes-sigs/cloud-provider-azure#current-status) correctly, it should be safe to use for version `v1.22` and higher. For v1.21 probably we should use `v1.0.5` of the image.

The out-of-tree CCM should basically be a superset of the in-tree. The in-tree only gets bug fixes since v1.20 and these are first implemented in the out-of-tree then ported back.

HappyTobi: Both versions work for 1.21 but the required parts for using the cluster on vmss-flex was also part of `v1.0.5`

HappyTobi: I will change `image.yaml` to use the `v1.0.5` for k8s `1.21`

HappyTobi: all tests are working with `v1.0.5` and k8s `1.21` I updated the pr

ialidzhikov (commented):
Thanks for the PR!

Basically I have the same question/concern about the backwards compatibility as in https://github.com/gardener/gardener-extension-provider-azure/pull/376#pullrequestreview-760489457. If we are sure that there are no backwards compatibility issues and if we tested thoroughly that our output qualification is passing against this change, then lgtm. Another option (IMO much safer one) would be to start using the out-of-tree CCM starting K8s v1.23 (or K8s v1.22) as we anyway have already the forks in-place for K8s v1.21 and K8s v1.22. I am just putting this option to the table because Azure itself supports several deployment models wrt to networking); the CCMs usually have number of provider-specific annotations to opt-in/opt-out special handlings and if the out-of-tree CCM code-base is a new one, then IMO gets a little risky to rollout such change all of a sudden. But as I said, if the change is well tested and compatibility is ensured, then lgtm from my side. 
If the only motivation behind this PR is to fix https://github.com/gardener/gardener-extension-provider-azure/issues/374, another option to tackle https://github.com/gardener/gardener-extension-provider-azure/issues/374 would be to revert the CCM version back to a working one until our upstream PR https://github.com/kubernetes/kubernetes/pull/105185 is merged, cherry-picked, released and adopted on our side.

> Also, can you please provide a proper release note?

Appropriate release note is still missing.

charts/internal/seed-controlplane/charts/cloud-controller-manager/templates/cloud-controller-manager.yaml
```diff
@@ -40,6 +40,8 @@ spec:
         {{- if semverCompare "< 1.17" .Values.kubernetesVersion }}
         - /hyperkube
         - cloud-controller-manager
+        {{- else if semverCompare ">=1.21" .Values.kubernetesVersion }}
```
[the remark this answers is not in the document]
HappyTobi: I will update that thx

HappyTobi:
Proper release notes:
```
Change "cloud-controller-manager" images for k8s versions >= 1.21.x
- 1.21.x changed to k8s sigs container version v.1.0.5
- 1.22 and above changed to k8s sigs container version v.1.2.2
For more information see https://github.com/kubernetes-sigs/cloud-provider-azure
```

kon-angelo:
After a discussion we had, we have decided to go with @ialidzhikov proposal and begin adoption with Kubernetes version v1.23. We will adapt it accordingly to the new versions once released, but until then the PR will remain open.

PR #406, opened 2021-11-30
Cleanup natipmigration logic

dkistner:
/kind cleanup
/platform azure

Remove `natgateway` ip migration (ref: `TODO(natipmigration)`) logic which was required to migrate from `azurerm` provider v1 to v2.
The logic is now in since `v1.15` and will be removed with `v1.23`.
This should be enough time to reconcile all `Infrastructure` resources. ref: https://github.com/gardener/gardener-extension-provider-azure/pull/192 

**Release note**:

```breaking operator
The logic to migrate the Terraform natgateway state will be removed. In case there are `Infrastructure` resources that are created with < `v1.15` and not reconciled with >= `v1.15` they need to be reconciled before a release with this change is applied.
```

/invite @kon-angelo 
/cc @ialidzhikov

### DIFF
--- a/hack/api-reference/api.md
+++ b/hack/api-reference/api.md
@@ -659,19 +659,6 @@ bool
 <p>Zoned indicates whether the cluster uses zones</p>
 </td>
 </tr>
-<tr>
-<td>
-<code>natGatewayPublicIpMigrated</code></br>
-<em>
-bool
-</em>
-</td>
-<td>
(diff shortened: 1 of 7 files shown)


kon-angelo (commented):
/lgtm 

/invite @ialidzhikov

kon-angelo (approved):

PR #410, opened 2021-12-06
Fixes deletion of remedy controller resources during cp migration

plkokanov:

/area control-plane-migration
/kind bug
/platform azure

This PR changes the migration logic of the controlplane resource so that first the controplane CRD charts are destroyed, thus deleting the remedy controller, afterwards finalizers on `PublicIPAddresses` and `VirtualMachines` are removed and finally the resources managed by the remedy controller are deleted.

This change is necessary now due to some recent changes which make the `PublicIPAddresses` and `VirtualMachines` be immediately recreated after being deleted (since the corresponding infrastructure resources still exist in the cluster), which was blocking the migration flow from completing.

Fixes #

/invite @stoyanr @dkistner 

**Release note**:

```other operator

```

### DIFF
--- a/pkg/controller/controlplane/actuator.go
+++ b/pkg/controller/controlplane/actuator.go
@@ -40,9 +40,6 @@ const (
 	GracefulDeletionTimeout = 10 * time.Minute
 )
 
-// TimeNow returns the current time. Exposed for testing.
-var TimeNow = time.Now
-
 // NewActuator creates a new Actuator that acts upon and updates the status of ControlPlane resources.
 func NewActuator(
 	a controlplane.Actuator,
@@ -80,7 +77,7 @@ func (a *actuator) InjectClient(client client.Client) error {
 
 // Delete reconciles the given controlplane and cluster, deleting the additional
 // control plane components as needed.
-// Before delegating to the composed Actuator, it ensures that all remedy controller resources have been deleted.
+// Before delegating to the composed Actuator, it ensures that all remedy controller resources have been deleted gracefully.
 func (a *actuator) Delete(
 	ctx context.Context,
 	cp *extensionsv1alpha1.ControlPlane,
@@ -92,13 +89,13 @@ func (a *actuator) Delete(
 			return err
 		}
 		if meta.LenList(list) != 0 {
-			if time.Now().Sub(cp.DeletionTimestamp.Time) <= a.gracefulDeletionTimeout {
+			if time.Since(cp.DeletionTimestamp.Time) <= a.gracefulDeletionTimeout {
 				a.logger.Info("Some publicipaddresses still exist. Deletion will be retried ...")
 				return &controllererror.RequeueAfterError{
 					RequeueAfter: a.gracefulDeletionWaitInterval,
 				}
 			} else {
-				a.logger.Info("Timeout while waiting for publicipaddresses to be gracefully deleted has expired. They will be forcefully removed.")
+				a.logger.Info("The timeout for waiting for publicipaddresses to be gracefully deleted has expired. They will be forcefully removed.")
 			}
 		}
 	}
@@ -166,12 +163,4 @@ func (a *actuator) forceDeleteRemedyControllerResources(ctx context.Context, cp 
 	}
 
 	return nil
-}
-
-func add(m map[string]string, key, value string) map[string]string {
-	if m == nil {
-		m = make(map[string]string)
-	}
-	m[key] = value
-	return m
 }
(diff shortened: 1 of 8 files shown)


dkistner:
Looks good to me.
/assign @stoyanr

pkg/controller/controlplane/actuator.go
```diff
@@ -70,14 +87,32 @@ func (a *actuator) Delete(
...
 			return err
 		}
+		if meta.LenList(list) != 0 {
+			if time.Now().Sub(cp.DeletionTimestamp.Time) <= a.gracefulDeletionTimeout {
+				a.logger.Info("Some publicipaddresses still exist. Deletion will be retried ...")
+				return &controllererror.RequeueAfterError{
+					RequeueAfter: a.gracefulDeletionWaitInterval,
+				}
+			} else {
+				a.logger.Info("Timeout while waiting for publicipaddresses to be gracefully deleted has expired. They will be forcefully removed.")
```
stoyanr, line 101:
```suggestion
				a.logger.Info("The timeout for waiting for publicipaddresses to be gracefully deleted has expired. They will be forcefully removed.")
```

pkg/controller/controlplane/actuator.go
```diff
@@ -17,37 +17,54 @@ package controlplane
...
+const (
+	// GracefulDeletionWaitInterval is the default interval for retry operations.
+	GracefulDeletionWaitInterval = 1 * time.Minute
+	// GracefulDeletionTimeout is the timeout that defines how long the actuator should wait for remedy controller resources to be deleted
+	// gracefully by the remedy controller itself
+	GracefulDeletionTimeout = 10 * time.Minute
+)
+
+// TimeNow returns the current time. Exposed for testing.
+var TimeNow = time.Now
```
stoyanr, line 44: This appears unused, is it really needed? If it is, the proper way to switch between "real time" and "testing time" would be via `clock.Clock` (could be added as a field to the actuator), see g/g code.

stoyanr (changes requested):
Looks quite good, I have only few minor comments. Will try to test it on my side before approving.

pkg/controller/controlplane/actuator.go
plkokanov: It is not needed.

stoyanr:
I tested migration and deletion with the latest changes. In both cases I manually created a `PublicIPAddress` to check the new functionality. Migration worked as expected, deletion waited for 10 minutes before eventually force-deleting the `PubilcIPAddress`. However, while waiting the reconciliation of controlplane happened constantly, instead of every 1 minute, even though the logs suggested that there is a requeue after of 1 minute:

```
{"level":"error","ts":"2021-12-08T08:23:16.974Z","logger":"controlplane_controller","msg":"error","namespace":"shoot--dev--i500152-az","name":"i500152-az","error":"Error deleting controlplane: requeue in 1m0s","stacktrace":"github.com/gardener/gardener/extensions/pkg/
{"level":"info","ts":"2021-12-08T08:23:16.985Z","logger":"controlplane_controller","msg":"Deleting the controlplane","namespace":"shoot--dev--i500152-az","name":"i500152-az"}
{"level":"info","ts":"2021-12-08T08:23:16.994Z","logger":"controlplane_controller","msg":"Starting the deletion of controlplane","controlplane":"shoot--dev--i500152-az/i500152-az"}
{"level":"info","ts":"2021-12-08T08:23:16.994Z","logger":"azure-controlplane-controller.azure-controlplane-actuator","msg":"Some publicipaddresses still exist. Deletion will be retried ..."}
{"level":"error","ts":"2021-12-08T08:23:16.994Z","logger":"controlplane_controller","msg":"error","namespace":"shoot--dev--i500152-az","name":"i500152-az","error":"Error deleting controlplane: requeue in 1m0s","stacktrace":"github.com/gardener/gardener/extensions/pkg/
{"level":"info","ts":"2021-12-08T08:23:17.005Z","logger":"controlplane_controller","msg":"Deleting the controlplane","namespace":"shoot--dev--i500152-az","name":"i500152-az"}
{"level":"info","ts":"2021-12-08T08:23:17.015Z","logger":"controlplane_controller","msg":"Starting the deletion of controlplane","controlplane":"shoot--dev--i500152-az/i500152-az"}
```

stoyanr:
We found the reason for the above behavior, it seems that the `RequeueAfterError` mechanism used here simply doesn't work with deletion (the resource is requeued immediately because a status update happens before that and the `IsDeleting` predicate is true). This is a general issue that should be fixed in g/g. Since it's not critical for our purposes here, we should simply merge this PR as it is. The erroneous behavior will go away once g/g is fixed and the fix is vendored.

stoyanr (approved):
/lgtm

PR #413, opened 2021-12-09
Transform deletionTimestamp to local time during graceful deletion check

plkokanov:

/area control-plane
/kind bug
/platform azure

During the deletion of the `ControlPlane` resource its deletionTimestamp will now be converted to local time when checking if the period to wait for `PublicIPAddress` resources to be gracefully deleted has elapsed.

Fixes #

/invite @dkistner @stoyanr 

**Release note**:

```other operator

```


plkokanov:
After looking into this further it seems that `Since()` is actually independent of timezones, so I will close this.

PR #423, opened 2022-01-03
Deploy allow-egress services for vmss flex clusters

dkistner:
/kind bug
/platform azure

Ensure that the `allow-egress` services are deployed for vmss flex (vmo) based clusters.
Otherwise it can happen that such a cluster does not have egress connectivity (in combination with `reverse-vpn` and no nat-gateway configured).

**Release note**:

```bugfix user
A bug has been fixed that could lead to no egress connectivity for a vmss flex (vmo) based Shoot clusters. This could happen if a cluster is using the `reverse-vpn` feature and has no nat-gateway configured.
```

/invite @kon-angelo


kon-angelo (dismissed):
/lgtm

kon-angelo (approved):
/lgtm

PR #430, opened 2022-01-24
Setting CSI Azure topology label on machine deployment creation to allow scale-up

himanshu-kun:
**How to categorize this PR?**

/area auto-scaling
/kind enhancement
/platform azure

**What this PR does / why we need it**:
adds driver specific CSI label `topology.disk.csi.azure.com/zone` during machineDeployment creation

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```other operator
machineDeployment will have the label `topology.disk.csi.azure.com/zone` when created if the shoot has csi enabled
```

PR #433, opened 2022-01-27
Include full zone name in topology label

himanshu-kun:

/area auto-scaling
/kind bug
/platform azure

This PR enhances the changes made by [this PR](https://github.com/gardener/gardener-extension-provider-azure/pull/430) to have full zone name rather than just the zone number as the label value.
Fixes #

**Release note**:

```other operator
NONE
```


himanshu-kun:
I have checked the behavior and it printing the entire name now. Since azure is different from other providers in terms of specifying the zone name, so I missed that.

kon-angelo (approved):
/lgtm

PR #450, opened 2022-02-25
Add integration test for user provided nat ips

dkistner:

/area testing
/kind enhancement
/platform azure

Add integration test to test the scenario where users provide own public ips for the Nat Gateway.

**Release note**:

```other operator
An integration test for user provided ip addresses for the NatGateway has been added.
```


kon-angelo (approved):
/lgtm

PR #465, opened 2022-03-15
Persist the infrastructure's ProviderStatus in the Infrastructure's state

plkokanov:

/area control-plane-migration
/kind bug
/platform azure

Persists the `Infrastructure.Status.ProviderStatus` also in the `Infrastructure.Status.State` field so that it can be saved in the `ShootState`. 
When restoring the infrastructure resource the ProviderStatus is loaded from the State field and added to the `Infrastructure.Status.ProviderStatus` so that the subsequent reconciliation can retrieve the necessary terraform values from it, instead of having to compute them.
The infrastructure resource must be reconciled at least once so that the state is updated, before migrations can be performed.

More info on why this is necessary is available in https://github.com/gardener/gardener/issues/4793

Fixes https://github.com/gardener/gardener/issues/4793

**Release note**:

```bugfix operator
Fixed an issue that prevented the migration of non-zoned clusters with availability sets, that occurs when the `CloudProfileConfig.DomainCounts` has changed and has different values compared to the ones used when creating the `Shoot`
```
```other operator
The `Infrastructure.Status.ProviderStatus` is now persisted in the `Infrastructure.Status.State` field after successful reconciliations. When the `Infrastructure` is restored the provider status is loaded from the `Infrastructure.Status.State` and added to the `Infrastructure.Status.ProviderStatus` at the start of the restore operation, so that any necessary terraform values can be retrieved from it instead of being computed.
```


pkg/internal/terraformfactory.go
```diff
@@ -0,0 +1,58 @@
...
+
+// NewTerraformer initializes a Terraformer
+func (f *terraformFactory) NewTerraformer(
+	logger logr.Logger,
+	restConfig *rest.Config,
+	purpose string,
+	infra *v1alpha1.Infrastructure,
+	useProjectedTokenMount bool,
+) (terraformer.Terraformer, error) {
+	return NewTerraformer(logger, restConfig, purpose, infra, useProjectedTokenMount)
```
dkistner, line 46: I guess this needs only for mocking purposes an exposed function right?

pkg/controller/infrastructure/actuator.go
```diff
@@ -31,16 +33,25 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+// State represents the last known State of an Infrastructure resource.
+// It is saved after a reconciliation and used during restore operations.
+type State struct {
```
dkistner, line 38: Just a nit, but maybe consider to rename this to `InfrastructureState` to make its purpose more clear.

pkg/controller/infrastructure/actuator_restore.go
```diff
@@ -26,9 +27,22 @@ import (
...
+		return err
+	}
+
+	terraformState, err := terraformer.UnmarshalRawState(infraState.TerraformState)
 	if err != nil {
 		return err
 	}
+
+	patch := client.MergeFrom(infra.DeepCopy())
+	infra.Status.ProviderStatus = infraState.SavedProviderStatus
```
dkistner, line 42: What will be the consequences if `infraState.SavedProviderStatus` is nil? e.g if the cluster is restored before once reconciled with this change?

plkokanov: The restoration will fail and the state will have to adapted manually to the new format. But I think this is ok since the control plane migration is still alpha.

plkokanov:
@dkistner I've applied the review comments, ptal.

pkg/controller/infrastructure/actuator.go
```diff
@@ -31,6 +32,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+// InfrastructureState represents the last known State of an Infrastructure resource.
+// It is saved after a reconciliation and used during restore operations.
+type InfrastructureState struct {
+	SavedProviderStatus *runtime.RawExtension `json:"savedProviderStatus,omitempty"`
```
dkistner, line 38: doc strings for exported properties. Sorry for missing that in my first review.

plkokanov: Sorry completely forgot about that as well. Add the comments now.

dkistner (approved):
/lgtm

PR #479, opened 2022-04-07
Assign value to `nodeTemplate.zone` in case of non-zonal setup

### LINKED ISSUE #476: MachhineClass is invalid: nodeTemplate.zone in body must be of type string: "null"
/area quality
/kind bug
/kind regression
/platform azure

**What happened**:
After https://github.com/gardener/gardener-extension-provider-azure/pull/417 Shoot fails to be created with:

```yaml
  lastOperation:
    description: 'Flow "Shoot cluster reconciliation" encountered task errors: [task
      "Waiting until shoot worker nodes have been reconciled" failed: Error while
      waiting for Worker shoot--foo--azure-local/azure-local to become ready:
      error during reconciliation: Error reconciling worker: failed to deploy the
      machine classes: failed to apply manifests: 1 error occurred: could not apply
      object of kind "MachineClass" "shoot--foo--azure-local/shoot--foo--azure-local-cpu-worker-7d27e":
      MachineClass.machine.sapcloud.io "shoot--foo--azure-local-cpu-worker-7d27e"
      is invalid: nodeTemplate.zone: Invalid value: "null": nodeTemplate.zone in body
      must be of type string: "null"]'
```

**What you expected to happen**:
Shoot creation to work as expected.

**How to reproduce it (as minimally and precisely as possible)**:
1. Create a non-zoned cluster

```yaml
    infrastructureConfig:
      kind: InfrastructureConfig
      networks:
        vnet:
          cidr: 10.222.0.0/16
        workers: 10.222.0.0/16
      zoned: false
```

2. Make sure that the creation fails with the above error.

**Anything else we need to know?**:

**Environment**:

- Gardener version (if relevant):
- Extension version: e9c2d2b110b58082fe0aa3449c2f107b7d653a8f
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- Others:

himanshu-kun:

/area auto-scaling
/kind regression
/platform azure

Earlier in case of non-zonal setup the `nodeTemplate.zone` field was left empty because of which it got converted to `null` string during marshalling to json, and failed CRD validation as the field is non-nullable.
Now `nodeTemplate.zone` is assigned value `no-zone` in case its a non-zonal setup.
The field value is used in CA to create a label during creation of template for node during scale-from-zero, and wouldn't affect any autoscaling.

Fixes #476 

**Release note**:

```other operator
NONE
```
/invite @kon-angelo @ialidzhikov

### DIFF
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -202,6 +202,7 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 			}
 
 			if pool.NodeTemplate != nil {
+				//TODO: make Zone field in nodeTemplate optional and not to pass it in case of non-zonal setup
 				zoneName := "no-zone"
 
 				if zone != nil {
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -202,7 +202,10 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 			}
 
 			if pool.NodeTemplate != nil {
-				//TODO: make Zone field in nodeTemplate optional and not to pass it in case of non-zonal setup
+				//	Currently Zone field is mandatory, and passing it an
+				//	empty string turns it to `null` string during marshalling which fails CRD validation
+				//	so setting it to a dummy value `no-zone`
+				//	TODO: Zone field in nodeTemplate optional and not to pass it in case of non-zonal setup
 				zoneName := "no-zone"
 
 				if zone != nil {
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -202,7 +202,7 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 			}
 
 			if pool.NodeTemplate != nil {
-				var zoneName string
+				zoneName := "no-zone"
 
 				if zone != nil {
 					zoneName = w.worker.Spec.Region + "-" + zone.name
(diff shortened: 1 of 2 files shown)


pkg/controller/worker/machines.go
```diff
@@ -202,7 +202,7 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 			}
 
 			if pool.NodeTemplate != nil {
-				var zoneName string
+				zoneName := "no-zone"
```
ialidzhikov, line 205: Wouldn’t the more sustainable approach be to make the field optional and simply omit it when the azure Shoot is not zoned one?

himanshu-kun: Yes I though abt that. The only issue is it would require MCM api change as `NodeTemplate` struct is there, and then changes in CA are required for validation, and revendoring new MCM release in all the extensions, which will delay scale-from-zero even further.
Since this is just Azure which faces this problem, so looking at the cost to benefit ratio, I took this approach.

ialidzhikov: Okay, I am fine with this change as a short-term mitigation. But in general it does not make sense to set dummy values for fields when they are not used but wrongly required in the CRD. In long-term I would suggest to open an issue about it, add a TODO here in provider-azure, and then do the required changes to make this field optional. WDYT?

himanshu-kun: Yes definitely, long term we need to make this optional. I'll do the required changes.

kon-angelo: Let's go with @ialidzhikov suggestion. Can you maybe add a comment/TODO to explain why we use the no-zone string and that it should be updated in the future ? Otherwise /lgtm

himanshu-kun: updated, I'll raise an issue in g/g now

kon-angelo (approved):
/lgtm

PR #482, opened 2022-04-12
Forbid k8s version upgrade from `< v1.21` to `>= v1.21.0`

dkistner:

/kind enhancement
/platform azure

As we are currently experience severe issues with the csi-driver for Azure, we will temporally forbid Shoot to upgrade from `< v1.21` to `>= v1.21.0` as with `v1.21` Azure clusters start to adopt csi.

This is a safety measure to not affect existing cluster with the current csi related issues.
The upgrade can still be enforced by adding the annotation `shoot.gardener.cloud/force-version-upgrade=true` to the Shoot on own risk.

New Shoot can still be created with `>= v1.21.0`.

**Release note**:

```other user
Azure Shoot cluster can temporally not be upgrade from Kubernetes version < v1.21 to >= v1.21.0. The upgrade can still be enforced by adding the annotation `shoot.gardener.cloud/force-version-upgrade=true` to the Shoot on own risk.
```

### DIFF
--- a/pkg/admission/validator/shoot.go
+++ b/pkg/admission/validator/shoot.go
@@ -179,6 +179,9 @@ func (s *shoot) validateUpdate(oldShoot, shoot *core.Shoot, cloudProfile *garden
 
 	allErrs = append(allErrs, s.validateShoot(shoot, oldInfraConfig, infraConfig, cloudProfile, cpConfig)...)
 
+	// TODO(dkistner) remove this once csi-driver async operation issues are resolved.
+	allErrs = append(allErrs, azurevalidation.ValidateUpgradeV120ToV121(shoot, oldShoot, specPath.Child("kubernetes", "version"))...)
+
 	return allErrs.ToAggregate()
 }
 
--- a/pkg/apis/azure/validation/shoot.go
+++ b/pkg/apis/azure/validation/shoot.go
@@ -31,6 +31,12 @@ import (
 
 const maxDataVolumeCount = 64
 
+var k8sV121 *semver.Version
+
+func init() {
+	k8sV121 = semver.MustParse("v1.21.0")
+}
+
 // ValidateNetworking validates the network settings of a Shoot.
 func ValidateNetworking(networking core.Networking, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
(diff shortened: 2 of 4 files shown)


pkg/azure/types.go
```diff
@@ -153,6 +153,10 @@ const (
 	// CSIMigrationKubernetesVersion is a constant for the Kubernetes version for which the Shoot's CSI migration will be
 	// performed.
 	CSIMigrationKubernetesVersion = "1.21"
+
+	// ForceK8sVersionUpgrade is an annotation set on the Shoot resource that allows to enforce a k8s version upgrade
+	// even if it would otherwise be forbidden by the admission controller.
+	ForceK8sVersionUpgrade = "shoot.gardener.cloud/force-version-upgrade"
```
kon-angelo, line 159: Just optional nit: I think it is more fitting  to `confirmation.gardener.cloud'`

pkg/apis/azure/validation/shoot.go
```diff
@@ -158,3 +164,28 @@ func validateVolumeFunc(volumeType *string, volumeSize string, encrypted *bool,
...
+	newShootVersion, err := semver.NewVersion(newShoot.Spec.Kubernetes.Version)
+	if err != nil {
+		return append(allErrs, field.Invalid(fldPath, newShoot.Spec.Kubernetes.Version, "could not parse kubernetes version"))
+	}
+
+	if oldShootVersion.LessThan(k8sV121) && (newShootVersion.Equal(k8sV121) || newShootVersion.GreaterThan(k8sV121)) {
+		if value, ok := newShoot.Annotations[azure.ForceK8sVersionUpgrade]; ok && value == "true" {
+			return allErrs
+		}
+		allErrs = append(allErrs, field.Forbidden(fldPath, "upgrade Kubernetes version from < v1.21 to >= v1.21.0 are disabled"))
```
kon-angelo, line 187: I think the error message should mention the workaround annotation. It is. more convenient for the user or maybe even the provider-extension issue for more information.

pkg/apis/azure/validation/shoot.go
```diff
@@ -31,6 +31,12 @@ import (
 
 const maxDataVolumeCount = 64
 
+var k8sV121 *semver.Version
+
+func init() {
+	k8sV121 = semver.MustParse("v1.21.0")
```
kon-angelo, line 37: Minor nit: I think it would be better if you would do this with `semver.NewConstraint("<= v1.21")`.

pkg/admission/validator/shoot.go
```diff
@@ -179,6 +179,9 @@ func (s *shoot) validateUpdate(oldShoot, shoot *core.Shoot, cloudProfile *garden
 
 	allErrs = append(allErrs, s.validateShoot(shoot, oldInfraConfig, infraConfig, cloudProfile, cpConfig)...)
 
+	// TODO(dkistner) remove this once csi-driver async operation issues are resolved.
+	allErrs = append(allErrs, azurevalidation.ValidateUpgradeV120ToV121(shoot, oldShoot, specPath.Child("kubernetes", "version"))...)
```
ialidzhikov, line 183: Probably it also makes sense to make this validation controlled by a flag/configuration option (enabled by default). This can be helpful for Gardener operators, they will be able to disable this new validation, if they need. WDYT?

ialidzhikov, line 183: We could also add a note to `docs/usage-as-end-user.md#csi-volume-provisioners` for the end users describing what is the issue/risk and that it is still possible to upgrade using the annotation `shoot.gardener.cloud/force-version-upgrade`.

kon-angelo: What would be the deprecation policy after we update to a CSI driver that fixes the issue ? 

Something like this ?
```
+1 version -> remove check internally
+2 versions -> remove flag
```

ialidzhikov: I don't have strong opinion. Yes, we can keep the validation in-place (disabled by default) until we gain experience with `azuredisk-csi-driver@v1.16.0`. Anyways, we can decide later how/when to proceed with removal of this validation/restriction.

ialidzhikov:
I believe we can close this PR after https://github.com/gardener/gardener-extension-provider-azure/pull/488 is merged (which should fix the observed issues during the upgrade from 1.20 to 1.21).

Issue #485, opened 2022-04-19
`Bastion` controller does not respect `InfrastructureStatus`


rfranzke:
**How to categorize this issue?**

/area ops-productivity
/kind bug
/platform azure

**What happened**:
Creating a `Bastion` for a `Shoot` with a non-default `InfrastructureConfig` does not work.
It seems the `Bastion` controller does not respect the `InfrastructureStatus` and instead makes assumptions of how the infrastructure layout of the cluster: https://github.com/gardener/gardener-extension-provider-azure/blob/07ac30778838935cdf19a73b2e15467896cf58ab/pkg/controller/bastion/options.go

**What you expected to happen**:
Creation of the `Bastion` should succeed.

**How to reproduce it (as minimally and precisely as possible)**:
1. Create a `Shoot` with

   ```yaml
   infrastructureConfig:
     apiVersion: azure.provider.extensions.gardener.cloud/v1alpha1
     kind: InfrastructureConfig
     networks:
       natGateway:
         enabled: true
       vnet:
         name: my-vnet
         resourceGroup: my-resource-group
       workers: 10.3.61.128/25
     zoned: true
   ```
2. Create a `Bastion` for this shoot cluster.

tedteng:
feel free to review it when you have a time thanks @rfranzke

tedteng:
new PR https://github.com/gardener/gardener-extension-provider-azure/pull/518

PR #492, opened 2022-04-27
fetch resource group, vNet, subnet value from Infra status

tedteng:

/area ops-productivity
/kind bug
/platform azure

fetch vNet, subnet value from Infra status, instead of joint name value
Fixes https://github.com/gardener/gardener-extension-provider-azure/issues/485

current solution logic depends on SingleSubnet
https://github.com/gardener/gardener-extension-provider-azure/blob/2ac7abfa4581d53fcc7bcfe98e51e46e2a60a5e7/pkg/controller/bastion/actuator_reconcile.go#L82-L84
**Release note**:

```other operator
fetch resource group, vNet, subnet value from Infra status, instead of joint name value
```

### DIFF
--- a/pkg/controller/bastion/actuator_reconcile.go
+++ b/pkg/controller/bastion/actuator_reconcile.go
@@ -21,12 +21,15 @@ import (
 	"net"
 	"time"
 
+	"github.com/gardener/gardener-extension-provider-azure/pkg/apis/azure"
+	"github.com/gardener/gardener-extension-provider-azure/pkg/apis/azure/helper"
 	azureclient "github.com/gardener/gardener-extension-provider-azure/pkg/azure/client"
 
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-05-01/network"
 	"github.com/gardener/gardener/extensions/pkg/controller"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
 	ctrlerror "github.com/gardener/gardener/pkg/controllerutils/reconciler"
+	"github.com/gardener/gardener/pkg/extensions"
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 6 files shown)


tedteng:
rebased as request, Please reveiw and process the PR when available @gardener/gardener-extension-provider-azure-maintainers Thanks

pkg/controller/bastion/actuator_reconcile.go
```diff
@@ -62,7 +70,15 @@ func (a *actuator) Reconcile(ctx context.Context, bastion *extensionsv1alpha1.Ba
 		return err
 	}
 
-	nic, err := ensureNic(ctx, factory, opt, publicIP)
+	if infrastructureStatus.Networks.Layout != "SingleSubnet" {
```
kon-angelo, line 73: Does it really not work when there are multiple subnets? I would think that the bastion instance would be able to reach the other nodes.

tedteng: I didn't test it, the main Azure bastion  PR base on a single Subnet mode is completed before multiple subnets are achieved https://github.com/gardener/gardener-extension-provider-azure/pull/331. but we have an internal ticket about `Bastion Azure extension multi zones support`

andrei-panov: Hi @kon-angelo, there are Firewall rules created together with Bastion that limit connections to the certain subnet. Originally there was a plan to implement multi-AZ right after the main Bastion development but priorities were changed, unfortunately...

andrei-panov: Just in case I dismiss our internal ticket because it's not visible to anyone...

tedteng: after rethinking, I think this PR also covers multiple subnets supported by a bastion extension now. By default, Azure subnets communicate with each other under the same virtual network.  https://docs.microsoft.com/en-us/azure/virtual-network/virtual-networks-udr-overview#:~:text=Azure%20automatically%20routes%20traffic,of%20a%20virtual%20network

I am assuming `infrastructureStatus.Networks.Subnets` always keep subnets which used by the gardener only,  by default bastion uses the first subnet to create a bastion
https://github.com/gardener/gardener-extension-provider-azure/blob/5297c61f6db5db460b719a6094169dbb580a01bb/pkg/controller/bastion/actuator_reconcile.go#L81

will remove the `SingleSubnet` mode check logic

tedteng: done https://github.com/gardener/gardener-extension-provider-azure/pull/492/commits/042d16bd153e5af48243d0b339062a6a754c5569

kon-angelo: @tedteng I checked it and as it is just removing the `SingleSubnet` check is not enough.
Here is a solution in a draft PR i opened based on this one. You can do some testing yourself too to see that it works E2E.

tedteng: Thanks, It works. I will close the PR. 
Please process the [draft PR ](https://github.com/gardener/gardener-extension-provider-azure/pull/518/commits/b7081b62b2d8f5687918b9a9ac48550219e39497)you opened

PR #519, opened 2022-06-09
User configurable node template

### LINKED ISSUE #501: Azure: Expected Capacity of machine should be able to be passed down to worker resource

/area autoscaling
/area usability
/kind enhancement
/kind api-change
/kind discussion

In a similar way to the already existing way to specify true machine capacity for AWS, we would like to have similar capability for Azure as well.
See: 
[(https://github.com/gardener/gardener/issues/4948)]
As well as:
[(https://github.com/gardener/gardener-extension-provider-aws/pull/461#issue-1078310771)]

**Why is this needed**:
In certain cases, for example when using vSMP MemoryONE virtualization technology, the actual resources/capacity of a machine, can be different than the ones originally reported by the hyper-scaler.
Hence, we would like to enable machine resource information to be passed down from cloudprofile to worker resource, and finally to machineclass. 
As the functionality already exists and works well on AWS, with our expansion to Azure, we would ask to provide Azure support for it as well.

himanshu-kun:

/area usability
/area robustness
/kind enhancement
/kind api-change
/platform azure

This PR makes node template user configurable. Users could provide their nodeTemplates under `spec.provider.worker.providerConfig` in shoot.yaml
Fixes #501 

- earlier providerConfig was unused for Azure , but now its used, so just like other extensions for aws and azure, type `workerConfig` is introduced.
**Release note**:

```user feature
Node template can now be configured from shoot spec under the path `spec.provider.worker.providerConfig` . Please refer the workerConfig section under docs `usage-as-end-user` for details
```


pkg/apis/azure/validation/worker.go
```diff
@@ -0,0 +1,65 @@
...
+	return allErrs
+}
+
+func validateNodeTemplate(nodeTemplate *extensionsv1alpha1.NodeTemplate, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if nodeTemplate == nil {
+		return nil
+	}
+	for _, capacityAttribute := range []corev1.ResourceName{"cpu", "gpu", "memory"} {
```
kon-angelo, line 45: why not use `corev1.ResourceCPU` etc for the resources instead ?

pkg/azure/client/mock/gomock_reflect_1503253645/prog.go
```diff
@@ -0,0 +1,72 @@
+package main
```
kon-angelo, line 1: I am pretty sure that this is a left-over of the mock generation process and should not exist/be commited. You can safely delete it.

kon-angelo:
Aside from the small nitpicks, I think its missing a couple validation tests, maybe also a test for the `machines.go` change but overall pretty good 👍

pkg/apis/azure/validation/worker.go
himanshu-kun: sure will do that

pkg/azure/client/mock/gomock_reflect_1503253645/prog.go
himanshu-kun: okay

himanshu-kun:
@kon-angelo I tried adding some validation tests but I felt like the existing cases are enough . Any suggestions are welcomed.

PR #521, opened 2022-06-14
Add PodSecurityPolicy for cloud-node-manager

ialidzhikov:
/area control-plane
/kind bug
/platform azure

With K8s 1.23 we enable OOT azure CCM. Currently for Shoots with `spec.kubenetes.allowPrivilegedContainers=false` the Shoot creation fails. The root cause is that cloud-node-manager is missing PSP and no cloud-node-manager Pods can be created:

```
$ k -n kube-system describe ds cloud-node-manager

Events:
  Type     Reason        Age                    From                  Message
  ----     ------        ----                   ----                  -------
  Warning  FailedCreate  15m (x79 over 15h)     daemonset-controller  Error creating: pods "cloud-node-manager-" is forbidden: PodSecurityPolicy: unable to admit pod: [spec.securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used]
  Warning  FailedCreate  2m5s (x16 over 4m49s)  daemonset-controller  Error creating: pods "cloud-node-manager-" is forbidden: PodSecurityPolicy: unable to admit pod: [spec.securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used]
```

**Release note**:

```bugfix user
An issue causing Shoot creation to fail for K8s > 1.23 clusters with `spec.kubenetes.allowPrivilegedContainers=false` is now fixed.
```

### DIFF
--- a/charts/internal/shoot-system-components/charts/cloud-controller-manager/templates/cloud-node-manager-psp-clusterrole.yaml
+++ b/charts/internal/shoot-system-components/charts/cloud-controller-manager/templates/cloud-node-manager-psp-clusterrole.yaml
@@ -1,0 +1,17 @@
+{{- if semverCompare ">= 1.23" .Capabilities.KubeVersion.GitVersion -}}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: gardener.cloud:psp:kube-system:cloud-node-manager
+rules:
+- apiGroups:
+  - policy
+  - extensions
+  resourceNames:
+  - extensions.gardener.cloud.provider-azure.cloud-node-manager
+  resources:
+  - podsecuritypolicies
+  verbs:
+  - use
+{{- end }}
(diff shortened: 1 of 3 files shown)


charts/internal/shoot-system-components/charts/cloud-controller-manager/templates/cloud-node-manager-psp-clusterrole.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if semverCompare ">= 1.23" .Capabilities.KubeVersion.GitVersion -}}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: gardener.cloud:psp:kube-system:cloud-node-manager
+rules:
+- apiGroups:
+  - policy
+  - extensions
```
kon-angelo, line 10: is the `extensions` also necessary ?

ialidzhikov: To my understanding K8s 1.15 can still serve the PodSecurityPolicy resource from the `extensions/v1beta1` API version. https://github.com/gardener/gardener-extension-provider-azure#supported-kubernetes-versions still lists 1.15 as technically supported K8s version. We may drop `extensions` as part of the clean up that drops support for the quite old K8s versions (1.15 and 1.16).

kon-angelo (approved):
/lgtm

PR #534, opened 2022-07-07
Do not allow containers in privileged mode for the cloud-node-manager PodSecurityPolicy

ialidzhikov:
/area security
/kind enhancement
/platform azure

There is no need to allow a container to run in privileged mode for the cloud-node-manager PodSecurityPolicy when there is no container currently in the cloud-node-manager that requires to run in privileged mode.

Follow-up after https://github.com/gardener/gardener-extension-provider-azure/pull/521
Indirectly part of https://github.com/gardener-security/standard-compliance/issues/3

**Release note**:

```other operator
NONE
```


kon-angelo (approved):
/lgtm

PR #535, opened 2022-07-12
Change default SC from HDD to SSD type

StenlyTU:

/area storage
/kind enhancement
/platform azure

Changing the default StorageClass for Azure shoots from Standard_LRS(HDD) to StandardSSD_LRS(SSD).

@dguendisch @hendrikKahl @BeckerMax 

**Release note**:

```other user
Changing the default StorageClass for Azure shoots from Standard_LRS(HDD) to StandardSSD_LRS(SSD)
```

### DIFF
--- a/charts/internal/shoot-storageclasses/templates/storageclasses.yaml
+++ b/charts/internal/shoot-storageclasses/templates/storageclasses.yaml
@@ -9,7 +9,7 @@ metadata:
     resources.gardener.cloud/delete-on-invalid-update: "true"
 provisioner: disk.csi.azure.com
 parameters:
-  skuName: Standard_LRS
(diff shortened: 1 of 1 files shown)


MSSedusch:
Azure still offers VM SKUs that do not support Premium Storage, even newer ones like the Dv5/Ev5, see https://docs.microsoft.com/azure/virtual-machines/dv5-dsv5-series
I assume you would have to do it similar to Accelerated networking where you store the information in the cloud profile.

dkistner (approved):
Thanks @MSSedusch then we can move on with `StandardSSD_LRS` as it is compatible with any machine type.

/lgtm

PR #544, opened 2022-07-22
Reenable the pod allocation for k8s <=1.23.x

kon-angelo:

/area control-plane
/kind bug
/platform azure

Caused by https://github.com/kubernetes/kubernetes/pull/97029 as the `allocate-node-cidrs` also controls route controller for `k8s < 1.21`

Fixes #

**Release note**:

```bugfix user
An issue preventing azure CCM to create routesfor K8s < 1.21 Shoot clusters is now fixed.
```


dkistner (approved):

ialidzhikov (commented):
Thanks for the PR!

Can you add a release note of type `bugfix user`? For example

```
An issue preventing azure CCM to create routesfor K8s < 1.21 Shoot clusters is now fixed.
```

An short explanation of the issue in the PR description would be also helpful for the readers.

Issue #555, opened 2022-08-09
`make verify` miss local helm


andrei-panov:
**How to categorize this issue?**

/area dev-productivity
/kind bug

**What happened**:
During execution of `make verify` it fails with missed `helm`

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

```Dockerfile
FROM ubuntu:xenial-20210804

RUN apt-get update && \
    apt-get install -y wget make git build-essential curl

RUN wget https://dl.google.com/go/go1.18.4.linux-amd64.tar.gz && \
    tar -C /usr/local -xzf go1.18.4.linux-amd64.tar.gz && \
    ln -s /usr/local/go/bin/go /usr/local/bin/go && \
    rm -rf go1.18.4.linux-amd64.tar.gz

#RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 && \
#    chmod 700 get_helm.sh && \
#    ./get_helm.sh

RUN go env

WORKDIR /root/go/src/github.com/gardener/gardener-extension-provider-azure
COPY . .
RUN make install
RUN make verify
```

**Anything else we need to know?**:
```
$ docker build --file Dockerfile2 .
Sending build context to Docker daemon  167.5MB
Step 1/8 : FROM ubuntu:xenial-20210804
 ---> b6f507652425
Step 2/8 : RUN apt-get update &&     apt-get install -y wget make git build-essential curl
 ---> Using cache
 ---> a8cb4e5a6cfc
Step 3/8 : RUN wget https://dl.google.com/go/go1.18.4.linux-amd64.tar.gz &&     tar -C /usr/local -xzf go1.18.4.linux-amd64.tar.gz &&     ln -s /usr/local/go/bin/go /usr/local/bin/go &&     rm -rf go1.18.4.linux-amd64.tar.gz
 ---> Using cache
 ---> b6e8bfaec581
Step 4/8 : RUN go env
 ---> Using cache
 ---> 3e8962f9d205
Step 5/8 : WORKDIR /root/go/src/github.com/gardener/gardener-extension-provider-azure
 ---> Running in 98877d862c60
Removing intermediate container 98877d862c60
 ---> ce868acd4cb4
Step 6/8 : COPY . .
 ---> c523b276db32
Step 7/8 : RUN make install
 ---> Running in 135682ce141a
> Install
Removing intermediate container 135682ce141a
[...]

PR #556, opened 2022-08-10
fix Makefile to locally install required helm

andrei-panov:

/area dev-productivity
/kind bug

install helm as required for `make verify`

Fixes #https://github.com/gardener/gardener-extension-provider-azure/issues/555


kon-angelo (approved):
/lgtm

Issue #551, opened 2022-08-01
Bastion controller fails to create Bastion because wrong resource group name is used to get the subnet for shoots with own vNet


vpnachev:
**How to categorize this issue?**

/area ops-productivity 
/kind bug
/platform azure

**What happened**:
For shoot clusters that brings their own vNet, the bastion controller fails to get the subnet as it uses the shoots resource group instead of the vNet resource group name.

In details, the subnet retrieval here https://github.com/gardener/gardener-extension-provider-azure/blob/61cfd0952e697b43a8bc06601358e836bcfdf9b7/pkg/controller/bastion/actuator.go#L196 is using the resource group name from opts which is set here https://github.com/gardener/gardener-extension-provider-azure/blob/61cfd0952e697b43a8bc06601358e836bcfdf9b7/pkg/controller/bastion/actuator_reconcile.go#L63 with the resource group where the shoot infrastructure resources are deployed.

However, when the shoot is using own vNet, it has its own resource group name which should be used in this case, i.e. `infrastructureStatus.Networks.VNet.ResourceGroup`.

I have prepared a quick patch that I think can resolve this issue, but I will need to test it.

<details>
  <summary>patch:</summary>

```patch
diff --git a/pkg/controller/bastion/actuator.go b/pkg/controller/bastion/actuator.go
index e7853410..39606d53 100644
--- a/pkg/controller/bastion/actuator.go
+++ b/pkg/controller/bastion/actuator.go
@@ -187,13 +187,18 @@ func getPublicIP(ctx context.Context, factory azureclient.Factory, opt *Options)
 	return ip, nil
 }
 
-func getSubnet(ctx context.Context, factory azureclient.Factory, vNet, subnetWork string, opt *Options) (*network.Subnet, error) {
+func getSubnet(ctx context.Context, factory azureclient.Factory, vNet, vNetResourceGroup, subnetWork string, opt *Options) (*network.Subnet, error) {
 	subnetClient, err := factory.Subnet(ctx, opt.SecretReference)
 	if err != nil {
 		return nil, err
 	}
 
-	subnet, err := subnetClient.Get(ctx, opt.ResourceGroupName, vNet, subnetWork, "")
+	rg := opt.ResourceGroupName
+	if vNetResourceGroup != "" {
+		rg = vNetResourceGroup
+	}
+
+	subnet, err := subnetClient.Get(ctx, rg, vNet, subnetWork, "")
 	if err != nil {
 		return nil, err
[...]

PR #560, opened 2022-08-15
bastion creation when own vnet adopted

tedteng:

/area ops-productivity
/kind bug
/platform azure

bastion resource group fetch fixed, when own vNet is adopted, the vNet resource group name is different from the default resource group name.

```
      kind: InfrastructureStatus
      networks:
        layout: SingleSubnet
        subnets:
        - name: shoot--sxxxa-xxx--azr-nodes
          purpose: nodes
        vnet:
          name: vnet-azr
          resourceGroup: rg-azr
      resourceGroup:
        name: shoot--sxxxa-xxx--azr
```
 
Fixes https://github.com/gardener/gardener-extension-provider-azure/issues/551

**Release note**:

```other operator
fix own vNet resource group name fetch in bastion creation
```
/invite @vpnachev

### DIFF
--- a/pkg/controller/bastion/actuator.go
+++ b/pkg/controller/bastion/actuator.go
@@ -185,12 +185,19 @@ func getPublicIP(ctx context.Context, log logr.Logger, factory azureclient.Facto
 }
 
 func getSubnet(ctx context.Context, log logr.Logger, factory azureclient.Factory, vNet, subnetWork string, opt *Options) (*network.Subnet, error) {
+	var sg string
 	subnetClient, err := factory.Subnet(ctx, opt.SecretReference)
 	if err != nil {
 		return nil, err
 	}
 
-	subnet, err := subnetClient.Get(ctx, opt.ResourceGroupName, vNet, subnetWork, "")
+	if opt.MyVnetEnabled {
+		sg = opt.MyVnetResourceGroupName
+	} else {
+		sg = opt.ResourceGroupName
+	}
+
+	subnet, err := subnetClient.Get(ctx, sg, vNet, subnetWork, "")
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/controller/bastion/actuator.go
+++ b/pkg/controller/bastion/actuator.go
@@ -21,15 +21,15 @@ import (
 	"encoding/json"
 	"fmt"
 
-	api "github.com/gardener/gardener-extension-provider-azure/pkg/apis/azure"
+	"github.com/gardener/gardener-extension-provider-azure/pkg/apis/azure"
 	azureclient "github.com/gardener/gardener-extension-provider-azure/pkg/azure/client"
-	"github.com/go-logr/logr"
-	"golang.org/x/crypto/ssh"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2021-03-01/compute"
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-05-01/network"
 	"github.com/gardener/gardener/extensions/pkg/controller"
 	"github.com/gardener/gardener/extensions/pkg/controller/bastion"
+	"github.com/go-logr/logr"
+	"golang.org/x/crypto/ssh"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/controller/bastion/options.go
```diff
@@ -79,19 +83,31 @@ func DetermineOptions(bastion *extensionsv1alpha1.Bastion, cluster *controller.C
...
 	}
-
+	infrastructureConfig := &api.InfrastructureConfig{}
+	if err = json.Unmarshal(cluster.Shoot.Spec.Provider.InfrastructureConfig.Raw, infrastructureConfig); err != nil {
+		return nil, err
+	}
+	// own vnet enabled if own vnet resourceGroup exist
+	myVnetEnabled := false
+	vNetrg := ""
+	if infrastructureConfig.Networks.VNet.ResourceGroup != nil && *infrastructureConfig.Networks.VNet.ResourceGroup != "" {
```
kon-angelo, line 93: I think a more appropriate way would to resolve the RG of the VNet is from the status e.g. [see here](https://github.com/gardener/gardener-extension-provider-azure/blob/d31fce706f474b8b1fe211dc22a882f3ba23a762/pkg/controller/worker/machines.go#L198)

pkg/controller/bastion/options.go
```diff
@@ -79,19 +83,31 @@ func DetermineOptions(bastion *extensionsv1alpha1.Bastion, cluster *controller.C
...
+		BastionInstanceName:     baseResourceName,
+		BastionPublicIPName:     publicIPResourceName(baseResourceName),
+		SecretReference:         secretReference,
+		CIDRs:                   cidrs,
+		WorkersCIDR:             workersCidr,
+		DiskName:                DiskResourceName(baseResourceName),
+		Location:                cluster.Shoot.Spec.Region,
+		ResourceGroupName:       resourceGroup,
+		MyVnetEnabled:           myVnetEnabled,
+		MyVnetResourceGroupName: vNetrg,
```
kon-angelo, line 107: I don't think that `MyVnetEnabled` is that useful as a field. I would do the following: 

Remove `MyVnetEnabled`

and

Rename `MyVnetResourceGroupName` -> `VnetResourceGroupName`. You can always calculate it when you calculate the `opt` and use that instead of introducing the conditionals in deep nested functions like `getSubnet`. 

You can alternatively, *not* introduce these 2 fields and perform the calculation in a place where you have the status, e.g. pass it with the other args in `ensureNic` [like here](https://github.com/gardener/gardener-extension-provider-azure/blob/d31fce706f474b8b1fe211dc22a882f3ba23a762/pkg/controller/bastion/actuator_reconcile.go#L74)

tedteng: done remove 2 fields and pass it in `ensureNic` bdd779563772700d8883ace9ceb911f121f52817

pkg/controller/bastion/options.go
tedteng: done remove 2 fields and pass it in `ensureNic` bdd779563772700d8883ace9ceb911f121f52817

tedteng:
This PR ready for review now

kon-angelo (approved):
/lgtm

pkg/controller/bastion/actuator.go
```diff
@@ -184,19 +184,26 @@ func getPublicIP(ctx context.Context, log logr.Logger, factory azureclient.Facto
...
 	}
 
-	subnet, err := subnetClient.Get(ctx, opt.ResourceGroupName, vNet, subnetWork, "")
+	if infrastructureStatus.Networks.VNet.ResourceGroup != nil {
+		sg = *infrastructureStatus.Networks.VNet.ResourceGroup
+	} else {
+		sg = opt.ResourceGroupName
+	}
+
+	subnet, err := subnetClient.Get(ctx, sg, infrastructureStatus.Networks.VNet.Name, infrastructureStatus.Networks.Subnets[0].Name, "")
```
petersutter, line 200: sorry for the late comment, I only got now aware of this PR as the corresponding issue was closed. Shouldn't the length of infrastructureStatus.Networks.Subnets be checked before? Or is it guaranteed at this point that it's not empty?

kon-angelo: It is done here: https://github.com/tedteng/gardener-extension-provider-azure/blob/bdd779563772700d8883ace9ceb911f121f52817/pkg/controller/bastion/actuator_reconcile.go#L145

petersutter: thanks @kon-angelo

PR #580, opened 2022-10-03
If machine image architecture is nil consider its architecture amd64

acumino:

/area control-plane
/kind bug
/platform azure

If a clusters gets stuck in the deletion before upgradation of  provider-azure extension version to a version which defaults the machineimage architecture field to `amd64` in worker status and if it uses a machine image that is removed from the cloud profile also then shoot deletion will complain that it can't delete shoot cause machine image is found neither in worker status nor in cloud profile of architecture `amd64`.
This PR enhances the `FindMachineImage` in such a way that if the machine image doesn't have an architecture field then it is set to `amd64`.

Fixes #

**Release note**:

```other operator
NONE
```


dkistner (approved):
good catch.
/lgtm

Issue #598, opened 2022-11-17
Handle inconsistent zone naming


timuthy:
**How to categorize this issue?**

/area high-availability
/area control-plane
/kind enhancement
/platform azure

**What would you like to be added**:
A new webhook should be added to the Azure provider extension that reacts on `Pod` creations and mutates node affinities that contain zone information.
The logic is supposed to prefix the zone by the shoot's region, like shown in the following example where the shoot is located in `westeurope`.

Input:

```
spec:
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: topology.kubernetes.io/zone
            operator: In
            values:
            - "2"
```

Output:

```
spec:
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: topology.kubernetes.io/zone
            operator: In
            values:
            - "westeurope-2"
```

**Why is this needed**:
Gardener expects a consistent zone naming across shoot and seed clusters.
When `ManagedSeed` workers are configured  `shoot.spec.provider.workers[].zones`, the very same information is replicated to `seed.spec.provider.zones`. This is furthermore used to add node affinities to control-plane pods in the seed (see [this document](https://github.com/gardener/gardener/blob/master/docs/development/high-availability.md) for more information).

Azure however, has an inconsistent naming in this matter and breaks the expectation. While zones are officially numeric only (`1`, `2`, `3`, ...), the CCM adds a compound of `region + zone` to the related node label, i.e. `topology.kubernetes.io/zone: {region}-{zone}`

Refs (thanks @kon-angelo):
- https://github.com/kubernetes-sigs/cloud-provider-azure/blob/f22ac733d38262af21e13d8b6215fc0cfc5a1633/pkg/provider/azure_zones.go#L163
- https://github.com/kubernetes-sigs/cloud-provider-azure/blob/f22ac733d38262af21e13d8b6215fc0cfc5a1633/pkg/provider/azure_zones.go#L124

PR #602, opened 2022-11-25
add topology webhook

kon-angelo:

/area control-plane
/kind bug
/platform azure

Introduce a pod mutation webhook for pods with inconsistent zone labeling due to CCM adding the zone label prefixed by the region.

Additionally increase the readiness initialDelay check because even with the selectors a pod webhook is slow on startup due to the initial synchronization.

Fixes https://github.com/gardener/gardener-extension-provider-azure/issues/598

**Release note**:

```other operator
Add topology webhook for pods with inconsistent zone labeling due to CCM adding the zone label prefixed by the region.
```


pkg/webhook/topology/mutator.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (m *mutator) InjectClient(client client.Client) error {
+	m.client = client
+	return nil
+}
+
+func (m *mutator) Mutate(ctx context.Context, new, _ client.Object) error {
+	if new == nil {
+		return nil
+	}
```
timuthy, line 51: I think you can skip this check as `nil` is not passed by the generic handler.

pkg/webhook/topology/mutator.go
```diff
@@ -0,0 +1,115 @@
...
+
+	"github.com/gardener/gardener-extension-provider-azure/pkg/azure"
+)
+
+type mutator struct {
+	client client.Client
+	log    logr.Logger
+}
+
+// New initializes a new mutator.
```
timuthy, line 36: Nit: Enhance doc string to tell what this mutator does.

pkg/webhook/topology/add.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/gardener/gardener-extension-provider-azure/pkg/azure"
+)
+
+const (
+	// WebhookName the name of the topology webhook.
+	WebhookName = "topology"
+	webhookPath = "topology"
+)
+
+var logger = log.Log.WithName("azure-topology-webhook")
```
timuthy, line 35:
```suggestion
var logger = log.Log.WithName("topology-webhook")
```

pkg/webhook/topology/mutator.go
```diff
@@ -0,0 +1,115 @@
...
+
+	newPod, ok := new.(*v1.Pod)
+	if !ok {
+		return fmt.Errorf("could not mutate: object is not type Pod")
+	}
+
+	gctx := gardenContext.NewGardenContext(m.client, new)
+	cluster, err := gctx.GetCluster(ctx)
+	if err != nil {
+		m.log.Error(err, "failed to mutate resource: %s/%s", new.GetNamespace(), new.GetName())
```
timuthy, line 66: This is no printf style logger, please use named fields instead. You can use the following example:

```
log := m.log.Error("pod", client.ObjectKeyFromObject(newPod))
...
log.Error(err, "failed to mutate resource")
```

pkg/webhook/topology/mutator.go
```diff
@@ -0,0 +1,115 @@
...
+		return err
+	}
+	return m.mutate(ctx, newPod, cluster)
+}
+
+func (m *mutator) mutate(_ context.Context, pod *v1.Pod, cluster *extensions.Cluster) error {
```
timuthy, line 72:
```suggestion
func (m *mutator) mutateNodeAffinity(pod *v1.Pod, cluster *extensions.Cluster) error {
```

timuthy (commented):
Thanks a lot for this PR. In general is looks good, for the rest please see my comments 🙂

pkg/webhook/topology/add.go
```diff
@@ -0,0 +1,69 @@
...
+		Path:     webhookPath,
+		Target:   extensionswebhook.TargetSeed,
+		Types:    types,
+		Webhook:  &admission.Webhook{Handler: handler, RecoverPanic: true},
+		Selector: &metav1.LabelSelector{
+			MatchLabels: map[string]string{
+				v1beta1constants.LabelSeedProvider: azure.Type,
+			},
+		},
+		ObjectSelector: &metav1.LabelSelector{
```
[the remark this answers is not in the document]
kon-angelo: We do have a namespace selector  at line 58 `v1beta1constants.LabelSeedProvider: azure.Type` in addition to the objectSelector. We can keep only the NS selector though if I understand correctly.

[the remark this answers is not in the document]
timuthy: I think it won't harm to add both to the namespace selector?!

pkg/webhook/topology/mutator.go
kon-angelo: Thank you for pointing this out. I proceeded differently and remove the logging because the webhook handler itself logs the errors so there is no need to do it a second time IMO.

pkg/webhook/topology/add.go
```diff
@@ -0,0 +1,65 @@
...
+
+const (
+	// WebhookName the name of the topology webhook.
+	WebhookName = "topology"
+	webhookPath = "topology"
+)
+
+var logger = log.Log.WithName("topology-webhook")
+
+// AddToManager creates a webhook adds the webhook to the manager.
```
timuthy, line 37:
```suggestion
// AddToManager creates a new topology webhook.
```
I saw that it's just copied but actually nothing is added to the manager here. We should rename these functions consistently in a separate PR.

pkg/webhook/topology/mutator.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/go-logr/logr"
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/gardener/gardener-extension-provider-azure/pkg/azure"
+)
+
+type mutator struct {
+	client client.Client
+	log    logr.Logger
```
timuthy, line 33: Can you remove the logger if it's not needed?

pkg/webhook/topology/mutator.go
```diff
@@ -0,0 +1,113 @@
...
+		return err
+	}
+	return m.mutateNodeAffinity(newPod, cluster)
+}
+
+func (m *mutator) mutateNodeAffinity(pod *v1.Pod, cluster *extensions.Cluster) error {
+	// probably can be omitted due to the namespace selector
+	if cluster.Seed.Spec.Provider.Type != azure.Type {
+		return nil
+	}
```
timuthy, line 74:
```suggestion
```

I also think it's safe to remove this check 👍

pkg/webhook/topology/add.go
```diff
@@ -0,0 +1,65 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package topology
+
+import (
+	extensionswebhook "github.com/gardener/gardener/extensions/pkg/webhook"
+	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+	v1 "k8s.io/api/core/v1"
```
rfranzke, line 20:
```suggestion
	corev1 "k8s.io/api/core/v1"
```

pkg/webhook/topology/mutator.go
```diff
@@ -0,0 +1,102 @@
...
+package topology
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	gardenContext "github.com/gardener/gardener/extensions/pkg/webhook/context"
+	"github.com/gardener/gardener/pkg/extensions"
+	v1 "k8s.io/api/core/v1"
```
rfranzke, line 24:
```suggestion
	corev1 "k8s.io/api/core/v1"
```

pkg/webhook/topology/mutator.go
```diff
@@ -0,0 +1,102 @@
...
+//  limitations under the License.
+
+package topology
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	gardenContext "github.com/gardener/gardener/extensions/pkg/webhook/context"
```
rfranzke, line 22: do not use camel case in import aliases

pkg/webhook/topology/add.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+var logger = log.Log.WithName("topology-webhook")
+
+// AddToManager creates a webhook adds the webhook to the manager.
+func AddToManager(mgr manager.Manager) (*extensionswebhook.Webhook, error) {
+	logger.Info("Adding webhook to manager")
+
+	types := []extensionswebhook.Type{
+		{Obj: &v1.Pod{}},
```
rfranzke, line 42: This will register a webhook for pods that reacts for both `CREATE` and `UPDATE` operations (ref: https://github.com/gardener/gardener/blob/219d828fcdea81fb3edf13de2736daf81e137923/extensions/pkg/webhook/registration.go#L290-L291). However, we only want to react on `CREATE` (these fields are anyways immutable on `UPDATE`). With this webhook in place, wouldn't we always run in errors when we try to update existing `Pod`s like this and the request gets rejected by the API server? Would it make sense to allow specifying the operations in the extension library for the webhook registration so that we only register the webhook for the `CREATE` rule?

kon-angelo: You can technically work around it I believe by not performing changes on requests where `old != nil`. That's not to say that the library shouldn't be enhanced just that it can be done async - not have it a prerequisite for this PR.

rfranzke: OK, but I don't see such check in the code. Can you also please open an issue in g/g and add a TODO statement here to remove this special handling once the (to-be-opened) g/g issue is fixed?

kon-angelo: > OK, but I don't see such check in the code.

Because I just became aware that it is needed :)  

> Can you also please open an issue in g/g and add a TODO statement here to remove this special handling once the (to-be-opened) g/g issue is fixed?

I added a TODO and a test. I will also open the ticket on g/g and assign myself but I would prefer if  for now we are not dependent on next gardener release for consumption. I think that anyway we would need to scan through our other webhooks after and remove this pattern for updates where it is used.

rfranzke (approved):

timuthy (approved):
/lgtm

PR #613, opened 2022-12-19
Fix topology webhook namespace lookup

kon-angelo:

/area bug
/kind control-plane
/platform azure

Refactors the topology webhook logic. Because of https://github.com/kubernetes/kubernetes/issues/88282 we can not depend on the pod namespace to find out cluster information on CREATE requests. So we refactor the handler to not use g/g library so that we have access to the request namespace and use that to retrieve the cluster. 

In addition we also try to fetch the region from the shoot-info in the case of managed-seeds as a backup.

Fixes #

**Release note**:

```other operator
NONE
```


MartinWeindel (approved):
/lgtm

PR #631, opened 2023-02-08
update namespace selector for topology webhook

kon-angelo:

/area control-plane
/kind bug
/platform azure

After https://github.com/gardener/gardener/pull/6997, handling only the shoot namespaces in the topology webhook is not enough. We should extend the applicable namespaces while excluding critical ones to prevent deadlocks.

Fixes #

**Release note**:

```breaking operator
Update the azure topology webhook to watch all namespaces and not just shoot namespaces. The `kube-system`, `garden` and extension namespaces are except to prevent deadlocks.
```

### DIFF
--- a/pkg/webhook/topology/add.go
+++ b/pkg/webhook/topology/add.go
@@ -58,7 +58,7 @@ func AddToManager(mgr manager.Manager) (*extensionswebhook.Webhook, error) {
 					Values:   []string{v1beta1constants.GardenRoleExtension},
 				},
 				{
-					Key:      "kubernetes.io/metadata.name",
+					Key:      corev1.LabelMetadataName
 					Operator: metav1.LabelSelectorOpNotIn,
 					Values:   []string{metav1.NamespaceSystem, v1beta1constants.GardenNamespace},
 				},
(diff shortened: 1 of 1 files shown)


MartinWeindel (approved):
/lgtm

pkg/webhook/topology/add.go
```diff
@@ -51,9 +51,17 @@ func AddToManager(mgr manager.Manager) (*extensionswebhook.Webhook, error) {
...
-				v1beta1constants.LabelSeedProvider: azure.Type,
-				v1beta1constants.GardenRole:        v1beta1constants.GardenRoleShoot,
+			MatchExpressions: []metav1.LabelSelectorRequirement{
+				{
+					Key:      v1beta1constants.GardenRole,
+					Operator: metav1.LabelSelectorOpNotIn,
+					Values:   []string{v1beta1constants.GardenRoleExtension},
+				},
+				{
+					Key:      "kubernetes.io/metadata.name",
```
timuthy, line 61:
```suggestion
					Key:      corev1.LabelMetadataName
```

timuthy (commented):
One nit. Can you please add a proper release note?

pkg/webhook/topology/add.go
kon-angelo: Ah, I tried to find the const for this but I failed. Thanks for the catch

timuthy:
Can you please check the failing `verify` step @kon-angelo?

timuthy (approved):
/lgtm
Thanks for taking care!

PR #668, opened 2023-04-14
alwasy using latest os version for bastion vm

tedteng:

/area ops-productivity
/kind enhancement
/platform Azure

It is similar to https://github.com/gardener/gardener-extension-provider-aws/issues/713, always search latest os version to build up the bastion instance instead of the hard code version.
Fixes #

**Release note**:

```other operator
always search latest os version to build up the Bastion instance
```

### DIFF
--- a/pkg/controller/bastion/actuator_reconcile.go
+++ b/pkg/controller/bastion/actuator_reconcile.go
@@ -22,7 +22,9 @@ import (
 	"regexp"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2021-03-01/compute"
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-05-01/network"
+	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/gardener/gardener/extensions/pkg/controller"
 	"github.com/gardener/gardener/extensions/pkg/util"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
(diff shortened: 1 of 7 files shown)


pkg/controller/bastion/actuator_reconcile.go
```diff
@@ -461,3 +469,32 @@ func findNextFreeNumber(set map[int32]bool, baseValue int32) *int32 {
...
+			sku.Insert(*v.Name)
+		}
+
+	}
+
+	if sku.List()[len(sku.List())-1] == "" {
+		return "", errors.New("sku not found")
+	}
+
+	return sku.List()[len(sku.List())-1], nil
```
kon-angelo, line 498: Maybe I am confused. Is the ordering - the "latest" as in the function name - implicit on the order of results returned by the API ? Because I see the filtering done here but not the ordering necessarily.

pkg/controller/bastion/actuator_reconcile.go
```diff
@@ -461,3 +469,32 @@ func findNextFreeNumber(set map[int32]bool, baseValue int32) *int32 {
 	set[baseValue] = true
 	return &baseValue
 }
+
+func getLatestSku(ctx context.Context, opt *Options, factory azureclient.Factory) (string, error) {
```
kon-angelo, line 473: I would honestly restructure the code a little bit here. Instead of "softcoding" the image by using the `vmImage` function directly, maybe modify the function signature so that it accepts a `compute.ImageReference`. So the function would indeed fetch the latest SKU for the passed image.

pkg/controller/bastion/actuator_reconcile.go
tedteng: yes it is  implicit on the order of results returned by the API as tested from my local for the `sku.List()` values
image

pkg/controller/bastion/actuator_reconcile.go
tedteng: @kon-angelo https://github.com/gardener/gardener-extension-provider-azure/pull/668/commits/a5ea43e5a18bb30feebe9982b9fca82f2fc572a8 committed, `vmImage` removed then I change to return `compute.ImageReference` from `getLatestSku`

pkg/controller/bastion/resourcesdefine.go
```diff
@@ -60,20 +60,15 @@ func publicIPAddressDefine(opt *Options) *network.PublicIPAddress {
...
 				VMSize: compute.VirtualMachineSizeTypesStandardB1s,
 			},
 			StorageProfile: &compute.StorageProfile{
-				ImageReference: &compute.ImageReference{
-					Publisher: to.StringPtr("Canonical"),
-					Offer:     to.StringPtr("UbuntuServer"),
-					Sku:       to.StringPtr("18.04-LTS"),
-					Version:   to.StringPtr("latest"),
-				},
+				ImageReference: vmImage(sku),
```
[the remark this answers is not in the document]
tedteng: @kon-angelo https://github.com/gardener/gardener-extension-provider-azure/pull/668/commits/a5ea43e5a18bb30feebe9982b9fca82f2fc572a8 committed.

kon-angelo (approved):
/lgtm

PR #669, opened 2023-04-14
Prevent shoot clusters with calico from configuring overlay network, which is not supported on azure

ScheererJ:

/area networking
/area usability
/kind bug
/platform azure

Prevent shoot clusters with calico from configuring overlay network, which is not supported on azure.

None.

/cc @DockToFuture 

**Release note**:

```other operator
Prevent shoot clusters from being configured with calico and overlay network as this is not supported on azure
```

### DIFF
--- a/pkg/admission/validator/shoot.go
+++ b/pkg/admission/validator/shoot.go
@@ -112,7 +112,7 @@ func (s *shoot) validateShoot(shoot *core.Shoot, oldInfraConfig, infraConfig *ap
 	allErrs := field.ErrorList{}
 
 	// Network validation
-	allErrs = append(allErrs, azurevalidation.ValidateNetworking(shoot.Spec.Networking, nwPath)...)
+	allErrs = append(allErrs, azurevalidation.ValidateNetworking(s.decoder, shoot.Spec.Networking, nwPath)...)
 
 	// Cloudprofile validation
 	allErrs = append(allErrs, azurevalidation.ValidateInfrastructureConfigAgainstCloudProfile(oldInfraConfig, infraConfig, shoot.Spec.Region, cloudProfile, infraConfigPath)...)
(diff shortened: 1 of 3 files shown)


pkg/admission/validator/shoot.go
```diff
@@ -113,6 +114,14 @@ func (s *shoot) validateShoot(shoot *core.Shoot, oldInfraConfig, infraConfig *ap
...
 	// Network validation
 	allErrs = append(allErrs, azurevalidation.ValidateNetworking(shoot.Spec.Networking, nwPath)...)
+	if shoot.Spec.Networking.Type == calico.ReleaseName {
+		networkConfig, err := decodeCalicoNetworkingConfig(s.decoder, shoot.Spec.Networking.ProviderConfig)
+		if err != nil {
+			allErrs = append(allErrs, field.InternalError(nwPath.Child("providerConfig"), err))
+		} else if networkConfig.Overlay != nil && networkConfig.Overlay.Enabled {
+			allErrs = append(allErrs, field.Forbidden(nwPath.Child("providerConfig").Child("overlay").Child("enabled"), "Calico overlay network is not supported on azure"))
+		}
+	}
```
kon-angelo, line 124: Can't it be part of the `validateNetworking` function (ln: 116) ?

kon-angelo (changes requested):

pkg/admission/validator/shoot.go
ScheererJ: Where should `decodeCalicoNetworkingConfig` go then? Should it be next to `validateNetworking`, where we do not have such functions as of now?

kon-angelo: You can probably use https://github.com/gardener/gardener-extension-provider-azure/blob/master/pkg/webhook/network/mutate.go#L36 but if not then we can move the function to this file or to a helper.go in the same package.

ScheererJ: Moved code as proposed. Please take another look.

ScheererJ:
Added tests as requested offline.

@kon-angelo Please take another look.

kon-angelo (approved):
/lgtm

PR #671, opened 2023-04-18
allow rg deletion with other resources

kon-angelo:

/area control-plane
/kind regression
/platform azure

Allow RG deletion even if it contains resources not managed by TF e.g. leftover PVs

Fixes #

**Release note**:

```other operator
Restore terraform behavior to delete the azure resource group even if it contains other resources.
```

### DIFF
--- a/pkg/internal/infrastructure/templates/main.tpl.tf
+++ b/pkg/internal/infrastructure/templates/main.tpl.tf
@@ -5,7 +5,11 @@ provider "azurerm" {
   client_secret   = var.CLIENT_SECRET
 
   skip_provider_registration = "true"
-  features {}
+  features {
+    resource_group {
+      prevent_deletion_if_contains_resources = false
+    }
+  }
 }
 
 #===============================================


pkg/internal/infrastructure/templates/main.tpl.tf
```diff
@@ -89,6 +93,8 @@ resource "azurerm_subnet" "{{ $workers }}" {
 {{- end }}
   address_prefixes          = ["{{ $subnet.cidr }}"]
   service_endpoints         = [{{ range $index, $serviceEndpoint := $subnet.serviceEndpoints }}{{ if $index }},{{ end }}"{{$serviceEndpoint}}"{{end}}]
+  private_link_service_network_policies_enabled = false
+  private_endpoint_network_policies_enabled = false
```
plkokanov, line 97: Was the default behaviour for these 2 fields also changed?

plkokanov (commented):
Generally looks good, I just have one question.

pkg/internal/infrastructure/templates/main.tpl.tf
kon-angelo: Ah, I was careless. Not this is a WIP commit from a different change I was preparing. Thank you for noticing

Issue #662, opened 2023-04-07
Add Infrastructure integration test for invalid credentials


ialidzhikov:
**How to categorize this issue?**

/area quality
/kind enhancement
/platform azure

**What would you like to be added**:
https://github.com/gardener/gardener-extension-provider-aws/pull/325 added such integration test for provider-aws. Verifying that the Infrastructure reconciliation with invalid credentials will lead to appropriate error code to be added to the Infrastructure status will also prevent issues like https://github.com/gardener/gardener-extension-provider-azure/issues/660.

**Why is this needed**:
To prevent potential issues with the error code matching/handling.

PR #692, opened 2023-05-29
Add integration test case for invalid credentials

tedteng:

/area usability
/kind test
/platform azure

This PR tests https://github.com/gardener/terraformer/pull/93 and https://github.com/gardener/gardener/pull/3950 by adding an integration test case for invalid credentials.
It creates an infrastructure resource but uses invalid credentials and ensures that the ERR_INFRA_UNAUTHENTICATED is correctly detected from the termination message and that the infrastructure can still be deleted afterwards (as it's a no-op).

Before this PR, the destroy logic and testing was executed as part of a cleanup action, while the destroy logic should be part of the actual test and not just cleanup. Also, this could lead to resource leakage on AliCloud and on the test-machinery cluster, because if one cleanup action fails (e.g. the destroy logic) no further cleanup actions are executed.
Now, the destroy logic is executed as a deferred function to ensure that the cleanup actions are still run, even if the test itself fails.
Fixes https://github.com/gardener/gardener-extension-provider-azure/issues/662

**Release note**:

```other operator
NONE
```
/invite @ialidzhikov

### DIFF
--- a/test/integration/infrastructure/infrastructure_test.go
+++ b/test/integration/infrastructure/infrastructure_test.go
@@ -17,6 +17,7 @@ package infrastructure_test
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
 	"path/filepath"
@@ -32,6 +33,7 @@ import (
 	azureRest "github.com/Azure/go-autorest/autorest/azure"
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+	gardencorev1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
 	"github.com/gardener/gardener/pkg/extensions"
 	"github.com/gardener/gardener/pkg/logger"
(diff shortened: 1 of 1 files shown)


tedteng:
/invite @ialidzhikov would you like to review this PR Thanks.

test/integration/infrastructure/infrastructure_test.go
```diff
@@ -489,6 +491,94 @@ var _ = Describe("Infrastructure tests", func() {
...
+			var (
+				namespace *corev1.Namespace
+				cluster   *extensionsv1alpha1.Cluster
+				infra     *extensionsv1alpha1.Infrastructure
+			)
+
+			framework.AddCleanupAction(func() {
+				By("cleaning up namespace and cluster")
+				Expect(client.IgnoreNotFound(c.Delete(ctx, namespace))).To(Succeed())
+				Expect(client.IgnoreNotFound(c.Delete(ctx, cluster))).To(Succeed())
```
kon-angelo, line 511: In the other tests, we have these in the defer part. Can I ask the reason to add these here instead ?

tedteng: The code I refer to https://github.com/gardener/gardener-extension-provider-aws/pull/325/files#diff-159fa2b5244aa8582f3ab173f88010e046ae979b06ef94e88b0346b1d308dc50R272-R278.  It seems we don't want to leak any resources even if defer not called.

tedteng: @kon-angelo

kon-angelo (approved):
/lgtm

PR #709, opened 2023-07-05
fix a case where a successful test would fail during cleanup

kon-angelo:

/area testing
/kind bug
/platform azure

Fix an 

Fixes #

**Release note**:

```other operator
Fix an issue with the bastion integration testing panicking on cleanup due to the security group having being deleted.
```


MartinWeindel (approved):
/lgtm

PR #716, opened 2023-08-04
Configure CCM to keep setting deprecated labes on the nodes

vpnachev:

/area ops-productivity robustness storage
/kind enhancement
/platform azure

Configure CCM to keep setting deprecated labes on the nodes for k8s clusters `>= 1.26, <1.28`. 

https://github.com/kubernetes-sigs/cloud-provider-azure/pull/2653 stopped maintaining the deprecated labels on the nodes, however some PVs are using these deprecated labels for node affinities and in the same time the PV node affinities field is immutable, hence once a shoot is upgraded to 1.26 it might happen that pods using exising PVs to not be scheduled on any node in the cluster.  

https://github.com/kubernetes-sigs/cloud-provider-azure/pull/3685 added the option the CCM to keep maintaining these labels, so let's make use of it. 
In the same time, k8s relaxed a bit the immutability the PV spec with https://github.com/kubernetes/kubernetes/pull/115391, but as of now this is available only in `>= 1.27` (there are opened cherry-picks to bring it also in 1.25 and 1.26). 

**Release note**:

```other user
The cloud-controller-manager is now set to keep setting deprecated node labels for k8s clusters of version `>=1.26.0, <1.28.0` to ensure pods using persistent volumes with node affinities are scheduled in the cluster.
```

### DIFF
--- a/charts/internal/seed-controlplane/charts/cloud-controller-manager/templates/cloud-controller-manager.yaml
+++ b/charts/internal/seed-controlplane/charts/cloud-controller-manager/templates/cloud-controller-manager.yaml
@@ -69,6 +69,9 @@ spec:
         - --tls-cipher-suites={{ .Values.tlsCipherSuites | join "," }}
         - --use-service-account-credentials
         - --v=2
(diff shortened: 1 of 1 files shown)


kon-angelo (approved):
/lgtm

PR #737, opened 2023-10-12
Vendor `gardener@1.81.0`, Upgrade dependencies and implement `ForceDelete`

acumino:

/kind enhancement
/area delivery open-source
/platform azure

- Vendor `gardener@1.81.0-dev`
- Implement `ForceDelete` method for extensions

Part of https://github.com/gardener/gardener/issues/3110

**Release note**:

```feature user
`provider-azure` extension now supports [Shoot Force Deletion](https://github.com/gardener/gardener/blob/master/docs/usage/shoot_operations.md#force-deletion). 
```


pkg/controller/controlplane/actuator.go
```diff
@@ -101,6 +101,16 @@ func (a *actuator) Delete(
...
 }
 
+// ForceDelete forcefully deletes the controlplane.
+func (a *actuator) ForceDelete(
+	ctx context.Context,
+	log logr.Logger,
+	cp *extensionsv1alpha1.ControlPlane,
+	cluster *extensionscontroller.Cluster,
+) error {
+	return a.Delete(ctx, log, cp, cluster)
```
plkokanov, line 111: WDYT about instead of calling the full `a.Delete()` function here to instead only call 
```go
        // Call Delete on the composed Actuator
	if err := a.Actuator.Delete(ctx, log, cp, cluster); err != nil {
		return err
	}
	if cp.Spec.Purpose == nil || *cp.Spec.Purpose == extensionsv1alpha1.Normal {
		// Delete all remaining remedy controller resources
		return a.forceDeleteRemedyControllerResources(ctx, log, cp)
	}
```
so that we avoid having to wait for the graceful deletion of the `PublicIPAddresses`?

kon-angelo: +1

PR #743, opened 2023-10-31
always set vmtype

kon-angelo:

/area control-plane
/kind technical-debt
/platform azure

Azure CCM's make a change to the default value of vmType in k8s v1.28. Instead of "reverting the conditions" where we set `vmType` let's always set the proper value.

With this change we will also always set the vmtype and not only in v1.28

Ref to azure PR: https://github.com/kubernetes-sigs/cloud-provider-azure/pull/4214

Fixes #

**Release note**:

```other operator
Always set the `vmType` in cloud-provider configuration files.
```


MartinWeindel (approved):
/lgtm

PR #761, opened 2023-12-11
Vendor gardener v1.83.2

kon-angelo:

/area control-plane
/kind enhancement
/platform azure

Fixes #

**Release note**:

```other dependency
Vendor gardener/gardener =>  `v1.83.2`
```


cmd/gardener-extension-provider-azure/app/app.go
```diff
@@ -202,8 +203,17 @@ func NewControllerManagerCommand(ctx context.Context) *cobra.Command {
 				return fmt.Errorf("could not update manager scheme: %w", err)
 			}
 
-			// add common meta types to schema for controller-runtime to use v1.ListOptions
-			metav1.AddToGroupVersion(scheme, machinev1alpha1.SchemeGroupVersion)
```
shafeeqes, line 206: Don't we still need this?

pkg/controller/healthcheck/add.go
```diff
@@ -113,14 +111,6 @@ func RegisterHealthChecks(ctx context.Context, mgr manager.Manager, opts healthc
 		workerConditionTypesToRemove = sets.New(gardencorev1beta1.ShootControlPlaneHealthy)
```
shafeeqes, line 111: These variables can be inlined now.

pkg/controller/infrastructure/infraflow/shared/basic_context.go
```diff
@@ -136,19 +136,13 @@ func (c *BasicFlowContext) AddTask(g *flow.Graph, name string, fn flow.TaskFn, o
...
 	if allOptions.Timeout > 0 {
 		tunedFn = tunedFn.Timeout(allOptions.Timeout)
 	}
 	task := flow.Task{
-		Name: name,
-		Fn:   c.wrapTaskFn(g.Name(), name, tunedFn),
-	}
+		Name:   name,
+		Fn:     c.wrapTaskFn(g.Name(), name, tunedFn),
+		SkipIf: allOptions.DoIf != nil && !*allOptions.DoIf}
```
shafeeqes, line 145: Nit
```suggestion
		SkipIf: allOptions.DoIf != nil && !*allOptions.DoIf
    }
```

pkg/controller/worker/machines_test.go
```diff
@@ -66,7 +65,7 @@ var _ = Describe("Machines", func() {
 		chartApplier = mockkubernetes.NewMockChartApplier(ctrl)
 		statusWriter = mockclient.NewMockStatusWriter(ctrl)
 
-		// Let the client always the mocked status writer when Status() is called.
+		// Let the seedClient always the mocked status writer when Status() is called.
```
shafeeqes, line 68:
```suggestion
		// Let the seed client always return the mocked status writer when Status() is called.
```
Similarly at other places.

cmd/gardener-extension-provider-azure/app/app.go
kon-angelo: I am actually a little bit confused about it but doesn't https://github.com/gardener/gardener-extension-provider-azure/blob/de1f4065099e7556d383cfe4787d162837ba2e12/cmd/gardener-extension-provider-azure/app/app.go#L198 already achieve the same ?

shafeeqes: That's true. Now, I'm not sure either. Maybe you can test it if it works this way, all the other extensions also can drop this lines then.

kon-angelo:
@shafeeqes PTAL :)

cmd/gardener-extension-provider-azure/app/app.go
kon-angelo: I didn't see any issue so far in my tests but i will keep an eye

pkg/controller/worker/machines_test.go
```diff
@@ -65,7 +65,7 @@ var _ = Describe("Machines", func() {
 		chartApplier = mockkubernetes.NewMockChartApplier(ctrl)
 		statusWriter = mockclient.NewMockStatusWriter(ctrl)
 
-		// Let the seedClient always the mocked status writer when Status() is called.
+		// Let the seed client always the mocked status writer when Status() is called.
```
shafeeqes, line 68: Nit
```
		// Let the seed client always return the mocked status writer when Status() is called.
```

shafeeqes (commented):
Thanks, just a small nit left.

pkg/controller/worker/machines_test.go
kon-angelo: Missed that one 👍

MartinWeindel (approved):
/lgtm

PR #787, opened 2024-02-05
simplify condition for picking terraform after trying to decode state

kon-angelo:

/area control-plane
/kind bug
/platform azure

The overall previous stricter check was causing issues in shoots that had not been reconciled for a long time and contained no state. The additional check done by decoding the state into the previous format, did not provide much benefit.

Fixes #

**Release note**:

```other operator
fix an issue where an empty infrastructure state would cause issues when picking the proper reconciler.
```


AndreasBurger (approved):
lgtm

PR #798, opened 2024-02-29
Disable remedy controller deployment

AndreasBurger:

/area control-plane
/kind cleanup
/platform azure

Scale the remedy-controller deployment down for existing clusters (on next reconciliation) and disable it for new clusters.

**Release note**:

```noteworthy operator
Disabled the deployment of the Remedy-Controller. Existing instances will be scaled down to zero on next reconciliation.
```


rfranzke:
Why don't we need it any longer? :)

kon-angelo:
@rfranzke We checked our (quite extensive) landscapes. The remedy controller is typically emitting prometheus metrics for attempted operations and we saw nothing for some time now. I take it that things are more stable on azure side.

The first phase is to disable the deployment of the remedy controller and see the effects if any. 
Aside from that, the remedy controller has 2 operations: 1 about reapplying failed VMs and the other about fixing orphan LB IPs. At least the first part about VMs, I reckon to be fitting for MCM provider-azure too but first let's confirm the impact of disabling the remedy controller for now.

rfranzke:
Thanks :) Just a thought: Does it make sense to make this configurable in order to be able to quickly re-enable it if necessary? But don't feel pushed, if it doesn't make much sense because the metrics indicate that it barely ever did something in the past weeks, it should be okay :)

PR #807, opened 2024-03-15
Adjust test parameters

hebelsan:

Increase test parameter timeout and update expected error code


kon-angelo (approved):
/lgtm

PR #824, opened 2024-04-05
Proposal: Data Disk restore from Image

### LINKED ISSUE #788: Integrate vSMP MemeoryOne in Azure
How to categorize this issue?

/area usability
/area os
/kind enhancement

What would you like to be added:
Gardener already supports multi-volume connection for machines used by MemeoryONE in AWS , to boot OS and MemeoryOne software on the same worker-node on separate dataVolumes.
The same functionality needed in Azure for MemoryONE worker-nodes.

I would post here a link to an epic for the similar functionality requested in AWS in the past.
https://github.com/gardener/gardener/issues/2354

Why is this needed:

elankath:

/area documentation
/kind enhancement
/platform azure

Fixes #788 

**Release note**:

```doc operator
NONE
```

### DIFF
--- a/docs/proposals/datadisk-image-restore.md
+++ b/docs/proposals/datadisk-image-restore.md
@@ -73,7 +73,7 @@ See [az image create](https://learn.microsoft.com/en-us/cli/azure/image?view=azu
 An optional `resourceGroup` can be specified if the image is associated with a non-default Azure resource group.
 
 The [MCM Azure Provider](https://github.com/gardener/machine-controller-manager-provider-azure) will ensure that the data 
-disk is created with the _image reference_ set to the provided `imgName`. See [az disk create](https://learn.microsoft.com/en-us/cli/azure/disk?view=azure-cli-latest#az-disk-create). 
+disk is created with the _image reference_ set to the provided `imgRef`. See [az disk create](https://learn.microsoft.com/en-us/cli/azure/disk?view=azure-cli-latest#az-disk-create). 
 The mechanics of this is left to MCM Azure provider.
 
 
--- a/docs/proposals/datadisk-image-restore.md
+++ b/docs/proposals/datadisk-image-restore.md
@@ -59,7 +59,7 @@ providerConfig:
   kind: WorkerConfig
   dataVolumes: # <-- NEW SUB_SECTION
   - name: vsmp1
-    imageName: imgName
+    imageReference: imgRef # ID or URN (publisher:offer:sku:version) of the image from which to create a disk
     resourceGroup: # (optional) Name of resource group. Will take default if omitted
   nodeTemplate: # (to be specified only if the node capacity would be different from cloudprofile info during runtime)
    capacity:
@@ -68,7 +68,7 @@ providerConfig:
      memory: 50Gi
 ```
 
-In the above `imgName` represents the image name created by an external process/tool.
+In the above `imgRef` specified via `providerConfig.dataVolumes.imageReference` represents the image ID or URN of the image from which to create a disk. This image should already exist.
 See [az image create](https://learn.microsoft.com/en-us/cli/azure/image?view=azure-cli-latest#az-image-create).
 An optional `resourceGroup` can be specified if the image is associated with a non-default Azure resource group.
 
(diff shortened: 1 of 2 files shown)


docs/proposals/datadisk-snapshot-restore.md
```diff
@@ -0,0 +1,79 @@
...
+
+In the above `snap-1234` represents the snapshot name created by an external process/tool.
+See [az-snapshot-create](https://learn.microsoft.com/en-us/cli/azure/snapshot?view=azure-cli-latest#az-snapshot-create).
+
+The Azure disk `snapshotName` is distinct from the azure `snapshotID`. The azure disk `snapshotID` is a full qualified hierarchical
+identifier that includes the `snapshotName`, Azure subscription ID and resource group name: 
+like `/subscriptions/<AzureSubscriptionID>/resourceGroups/<resourceGroupName>/providers/Microsoft.Compute/disks/<snapshotName>`
+
+It would be painful and errorprone to specify this in the shoot `WorkerConfig` section, so it is best that the MCM Azure provider take care
+of forming the fully qualified azure disk `snapshotID` and forming the `MachineClass` for azure which is then operated on by MCM Azure Provider.
```
kon-angelo, line 77: Two points:
a) I am not sure why you consider error prone the copy-pasting or an azure UID compared to MCM constructing. That being said it is a trick used often by the azure controllers so we can do it too, but I don't necessarily find it easier.
b) Main issue is that you need at least to also add the resource group to the spec. Otherwise you are constrained to using disks only in the same resource group and that would make it useless for the vsmp scenario, as I imagine they want to restore from some knowd resource in a RG different from the shoots.

docs/proposals/datadisk-snapshot-restore.md
```diff
@@ -0,0 +1,79 @@
...
+    - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+- [Alternatives](#alternatives)
+
+## Summary
+
+Currently, we have no support either in the shoot spec or in the [MCM Azure](https://github.com/gardener/machine-controller-manager-provider-azure) for restoring Azure Data Disks from snapshots
+
+## Motivation
+The primary motivation is to support [Integration of vSMP MemeoryOne in Azure #](https://github.com/gardener/gardener-extension-provider-azure/issues/788).
```
kon-angelo, line 29: I think that the proposal can stand on its own without any reference to vsmp.

kon-angelo (changes requested):

docs/proposals/datadisk-snapshot-restore.md
elankath: a) I removed this statement.
b) added `resourceGroup` as an optional field, though I am unsure whether this is needed for vsmp case. They just appear to use the default resource group. It appears `-g default` is specified everywhere in their script.

docs/proposals/datadisk-image-restore.md
```diff
@@ -0,0 +1,79 @@
...
+```
+We propose that the worker config section be enahnced to support data disk configuration
+```yaml
+providerConfig:
+  apiVersion: azure.provider.extensions.gardener.cloud/v1alpha1
+  kind: WorkerConfig
+  dataVolumes: # <-- NEW SUB_SECTION
+  - name: vsmp1
+    imageReference: imgRef # ID or URN (publisher:offer:sku:version) of the image from which to create a disk
+    resourceGroup: # (optional) Name of resource group. Will take default if omitted
```
kon-angelo, line 63: What is the resource group supposed to do ? I think you don't need to reference a resource group to consume an image right ?

docs/proposals/datadisk-image-restore.md
```diff
@@ -0,0 +1,79 @@
...
+      memory: 50Gi
+```
+We propose that the worker config section be enahnced to support data disk configuration
+```yaml
+providerConfig:
+  apiVersion: azure.provider.extensions.gardener.cloud/v1alpha1
+  kind: WorkerConfig
+  dataVolumes: # <-- NEW SUB_SECTION
+  - name: vsmp1
+    imageReference: imgRef # ID or URN (publisher:offer:sku:version) of the image from which to create a disk
```
kon-angelo, line 62: So is the plan for MCM to parse the ID provided and resolve it to an image? MCM-azure already has a struct to ref these in https://github.com/gardener/machine-controller-manager-provider-azure/blob/e86af1fe835f427b1d16b4316b3f02752bd7cd73/pkg/azure/api/providerspec.go#L145 . Why not use that instead ?

kon-angelo (changes requested):

PR #833, opened 2024-04-17
Switch to new SDK clients and update deps

AndreasBurger:

/area quality
/kind task
/platform azure

Incorporates the Upgrade PR proposed by dependabot and updates most clients of the old SDK to the new one.


kon-angelo (approved):
/lgtm

PR #843, opened 2024-04-25
delete empty RGs on infra creation

kon-angelo:

/area control-plane
/kind bug
/platform azure

Due to the Terraform/Azure interaction, there may be resource groups that are created but not in the TF state, thus blocking infra creation

Fixes #

**Release note**:

```other operator
Delete empty resource groups on infrastructure creation. 
```

### DIFF
--- a/pkg/apis/azure/helper/scheme.go
+++ b/pkg/apis/azure/helper/scheme.go
@@ -131,9 +131,7 @@ func InfrastructureStateFromRaw(raw *runtime.RawExtension) (*api.InfrastructureS
 func InfrastructureStatusFromInfrastructure(infra *extensionsv1alpha1.Infrastructure) (*api.InfrastructureStatus, error) {
 	status := &api.InfrastructureStatus{}
 	if infra.Status.ProviderStatus != nil && infra.Status.ProviderStatus.Raw != nil {
-		if _, _, err := lenientDecoder.Decode(infra.Status.ProviderStatus.Raw, nil, status); err != nil {
-			return nil, err
-		}
+		return InfrastructureStatusFromRaw(infra.Status.ProviderStatus)
 	}
 	return status, nil
 }
--- a/pkg/apis/azure/helper/scheme.go
+++ b/pkg/apis/azure/helper/scheme.go
@@ -60,12 +60,12 @@ func InfrastructureConfigFromInfrastructure(infra *extensionsv1alpha1.Infrastruc
 // InfrastructureStatusFromRaw extracts the InfrastructureStatus from the
 // ProviderStatus section of the given Infrastructure.
 func InfrastructureStatusFromRaw(raw *runtime.RawExtension) (*api.InfrastructureStatus, error) {
-	config := &api.InfrastructureStatus{}
+	status := &api.InfrastructureStatus{}
 	if raw != nil && raw.Raw != nil {
-		if _, _, err := lenientDecoder.Decode(raw.Raw, nil, config); err != nil {
+		if _, _, err := lenientDecoder.Decode(raw.Raw, nil, status); err != nil {
 			return nil, err
 		}
-		return config, nil
+		return status, nil
 	}
 	return nil, fmt.Errorf("provider status is not set on the infrastructure resource")
 }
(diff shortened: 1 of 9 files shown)


pkg/azure/client/resources.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// NewResourceClient creates a new ResourceClient
+func NewResourceClient(auth *internal.ClientAuth, tc azcore.TokenCredential, opts *arm.ClientOptions) (*ResourceClient, error) {
+	client, err := armresources.NewClient(auth.SubscriptionID, tc, opts)
+	return &ResourceClient{client: client}, err
+}
+
+// ListByResourceGroup fetches all resources of a resource group.
+func (c *ResourceClient) ListByResourceGroup(ctx context.Context, resourceGroupName string) ([]*armresources.GenericResourceExpanded, error) {
```
timuthy, line 31: Does it make sense to optionally pass `armresources.ClientListByResourceGroupOptions`? Background: We can set `Top: 1` to save bandwidth and memory when checking for an empty resource group.

pkg/controller/infrastructure/terraform_reconciler.go
```diff
@@ -85,6 +87,10 @@ func (r *TerraformReconciler) reconcile(ctx context.Context, infra *extensionsv1
 		return err
 	}
 
+	if err := r.cleanResourceGroupIfNeeded(ctx, infra, cluster, cfg); err != nil {
+		return err
+	}
+
```
timuthy, line 93: How about only executing this when an apply error happens?

timuthy (commented):
Thanks a lot for the workaround 🚀

timuthy:
Can you also add more context to the release note, so that people are aware of what this is good for?

pkg/apis/azure/helper/scheme.go
```diff
@@ -124,3 +124,16 @@ func InfrastructureStateFromRaw(raw *runtime.RawExtension) (*api.InfrastructureS
...
 	return state, nil
 }
+
+// InfrastructureStatusFromInfrastructure extracts the InfrastructureStatus from the
+// ProviderStatus section of the given Infrastructure.
+// If the providerConfig is missing from the status, it will return a zero-value InfrastructureStatus.
+func InfrastructureStatusFromInfrastructure(infra *extensionsv1alpha1.Infrastructure) (*api.InfrastructureStatus, error) {
+	status := &api.InfrastructureStatus{}
+	if infra.Status.ProviderStatus != nil && infra.Status.ProviderStatus.Raw != nil {
+		if _, _, err := lenientDecoder.Decode(infra.Status.ProviderStatus.Raw, nil, status); err != nil {
```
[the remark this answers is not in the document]
kon-angelo: Is the question about using something like:
```
func InfrastructureStatusFromInfrastructure(infra *extensionsv1alpha1.Infrastructure) (*api.InfrastructureStatus, error) {
	status := &api.InfrastructureStatus{}
	if infra.Status.ProviderStatus != nil && infra.Status.ProviderStatus.Raw != nil {
		return InfrastructureStatusFromRaw(infra.Status.ProviderStatus)
	}
	return status, nil
}
```

or why do I needed to implement a "similar" function"

pkg/controller/infrastructure/terraform_reconciler.go
```diff
@@ -234,3 +216,79 @@ func (r *TerraformReconciler) Delete(ctx context.Context, infra *extensionsv1alp
...
+	} else if !ok {
+		return nil
+	}
+
+	resourceClient, err := clientFactory.Resource()
+	if err != nil {
+		return err
+	}
+	res, err := resourceClient.ListByResourceGroup(ctx, rgName)
+	if client.IgnoreNotFound(err) != nil {
```
AndreasBurger, line 283: Thats an interesting case - the resource group being absent after we explicitly checked for its existence. I agree that it's not worth to make a fuss about it at that point, but it might be worth a log-line if that case occurs.

pkg/controller/infrastructure/terraform_reconciler.go
```diff
@@ -94,6 +98,29 @@ func (r *TerraformReconciler) reconcile(ctx context.Context, infra *extensionsv1
...
+			"error", err)
+
+		ok, inErr := r.cleanResourceGroupIfNeeded(ctx, infra, cluster, cfg)
+		if inErr == nil && ok {
+			// we return a retryable error for the controller to retry instead of locking the user to a non-retryable error.
+			return gardencorev1beta1helper.NewErrorWithCodes(fmt.Errorf("retry after resource group cleanup"), gardencorev1beta1.ErrorRetryableInfraDependencies)
+		}
+		if inErr != nil {
+			r.Logger.Error(inErr, "checking and cleaning up the resource group after a failed TF apply failed")
+		}
```
AndreasBurger, line 122: Consider sth. with a little bit more structure like 
```golang
if ok {
  …
} else {
 if err != nil {
  …
  }
}
```

AndreasBurger (approved):
lgtm

timuthy (approved):
/lgtm

PR #844, opened 2024-04-25
Backport of PR 843 to release-branch

AndreasBurger:

/area control-plane
/kind bug
/platform azure

Due to the Terraform/Azure interaction, there may be resource groups that are created but not in the TF state, thus blocking infra creation

Fixes #

**Release note**:

```other operator
The extension will now try to delete empty resource groups on infrastructure creation after an unsuccessful terraform-apply operation. 
A resource group may not be ready for some time after a successful create call returns. The azurerm terraform-provider on resource group does not respect that and the GET call may result in a NotFound error creating a deadlock. The extension will try to workaround this by deleting empty resource groups under the condition that this is a Create operation.
```

### DIFF
--- a/pkg/controller/infrastructure/terraform_reconciler_test.go
+++ b/pkg/controller/infrastructure/terraform_reconciler_test.go
@@ -70,6 +70,9 @@ var _ = Describe("Actuator", func() {
 		tfState            *terraformer.RawState
 		revert             func()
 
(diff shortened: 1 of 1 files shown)


timuthy (approved):
/lgtm

kon-angelo (approved):
/lgtm

PR #851, opened 2024-05-02
fix nil pointer for buckets

kon-angelo:

/area control-plane
/kind bug
/platform azure

Fix an issue where the buckup bucket client was not properly created.

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/controller/backupbucket/actuator.go
+++ b/pkg/controller/backupbucket/actuator.go
@@ -36,18 +36,12 @@ func newActuator(mgr manager.Manager) backupbucket.Actuator {
 }
 
 func (a *actuator) Reconcile(ctx context.Context, _ logr.Logger, backupBucket *extensionsv1alpha1.BackupBucket) error {
-
 	backupConfig, err := helper.BackupConfigFromBackupBucket(backupBucket)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


pkg/controller/backupbucket/actuator.go
```diff
@@ -36,18 +36,12 @@ func newActuator(mgr manager.Manager) backupbucket.Actuator {
...
 		return err
 	}
 
-	var cloudConfiguration *azure.CloudConfiguration
-	if backupBucket != nil {
-		cloudConfiguration = backupConfig.CloudConfiguration
-	}
-
-	azCloudConfiguration, err := azureclient.AzureCloudConfigurationFromCloudConfiguration(cloudConfiguration)
+	azCloudConfiguration, err := azureclient.AzureCloudConfigurationFromCloudConfiguration(backupConfig.CloudConfiguration)
```
hebelsan, line 44: Is this really safe?
I mean backupConfig is still a pointer.
For this reason I would prefer that helpers like `BackupConfigFromBackupBucket` return a struct value instead of a pointer.

hebelsan (approved):

AndreasBurger (approved):

PR #852, opened 2024-05-03
Provide Egress CIDRs in Infra Status

AndreasBurger:

/area monitoring
/kind enhancement
/platform azure

With this PR the CIDR-Ranges used for cluster-egress will be provided in the shoot's infrastructure resource.

Fixes #

**Release note**:

```other operator
The CIDR blocks used for shoot egress will now be provided via the status of the shoot's infrastructure-resource
```

### DIFF
--- a/pkg/controller/infrastructure/infraflow/ensurer.go
+++ b/pkg/controller/infrastructure/infraflow/ensurer.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v5"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v4"
@@ -512,6 +513,9 @@ func (fctx *FlowContext) ensureNatGateways(ctx context.Context) error {
 		return joinError
 	}
 
+	ipClient, _ := fctx.factory.PublicIP()
+	ipAddresses := []string{}
+
 	for name, nat := range toReconcile {
 		nat, err := c.CreateOrUpdate(ctx, fctx.adapter.ResourceGroupName(), name, *nat)
 		if err != nil {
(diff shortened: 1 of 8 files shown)


pkg/controller/infrastructure/actuator_helper.go
```diff
@@ -25,10 +25,14 @@ func patchProviderStatusAndState(
 	infra *extensionsv1alpha1.Infrastructure,
 	status *v1alpha1.InfrastructureStatus,
 	state *runtime.RawExtension,
+	egressCidrs *[]string,
```
hebelsan, line 28: Passing egressCidrs as pointer slice is imp not necessary.
Just pass an empty slice if it's not set.

pkg/controller/infrastructure/infraflow/ensurer.go
```diff
@@ -739,6 +757,22 @@ func (f *FlowContext) GetInfrastructureState() (*runtime.RawExtension, error) {
 	}, nil
 }
 
+// GetEgressIpCidrs retrieves the CIDRs of the IP ranges used for egress from the FlowContext
+func (f *FlowContext) GetEgressIpCidrs() *[]string {
```
hebelsan, line 761: again no need to return a pointer to slice - you can return nil also for slices

pkg/controller/infrastructure/terraform_reconciler_test.go
```diff
@@ -165,6 +165,7 @@ var _ = Describe("Actuator", func() {
 			tf.EXPECT().Apply(ctx)
 			tf.EXPECT().GetStateOutputVariables(ctx, gomock.Any())
 			tf.EXPECT().GetRawState(ctx).Return(tfState, nil)
+			tf.EXPECT().GetRawState(ctx).Return(tfState, nil)
```
hebelsan, line 168: TODO?

hebelsan (changes requested):

pkg/controller/infrastructure/terraform_reconciler_test.go
AndreasBurger: No, this was due to `GetRawState` being called twice. Admittedly, that was not the ideal solution.

I've moved the call to `GetRawState` up and passed the result to the functions reducing it back to one expected call.

AndreasBurger:
Seeing some unrelated errors (that also differ from what I see locally) so
/test

kon-angelo:
Looks good but

/needs rebase

kon-angelo (approved):
/lgtm

GL with the tests 😉

AndreasBurger:
fixed nil-ptr deref
/test

hebelsan (approved):

PR #875, opened 2024-05-29
Detect Cloud Instance by Region as Fallback

AndreasBurger:
/area usability
/kind enhancement
/platform azure

With the changes in this PR, the extension will look at the name of the region configured in the shoot to determine the correct cloud instance to connect to iff there is no explicit CloudConfiguration provided.

**Release note**:

```noteworthy user
When no CloudConfiguration is given, the Azure Cloud instance to connect to will be inferred from the region configured for the shoot
```

### DIFF
--- a/pkg/apis/azure/helper/helper.go
+++ b/pkg/apis/azure/helper/helper.go
@@ -6,7 +6,6 @@ package helper
 
 import (
 	"fmt"
-	"strings"
 
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	"k8s.io/utils/ptr"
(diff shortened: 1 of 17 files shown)


pkg/apis/azure/helper/helper.go
```diff
@@ -145,3 +146,24 @@ func InfrastructureZoneToString(zone int32) string {
...
+
+// CloudInstanceNameFromRegion returns the canonical cloud instance name for the instance the region is hosted in.
+func CloudInstanceNameFromRegion(region string) string {
+	if hasAnyPrefix(region, api.AzureGovRegionPrefixes...) {
+		return api.AzureGovCloudName
+	} else if hasAnyPrefix(region, api.AzureChinaRegionPrefixes...) {
+		return api.AzureChinaCloudName
+	} else {
+		return api.AzurePublicCloudName
+	}
```
kon-angelo, line 158: Maybe a switch statement would look cleaner here

kon-angelo: Also I wouldn't necessarily put this in the api/helper package

pkg/controller/backupbucket/actuator.go
```diff
@@ -41,7 +41,13 @@ func (a *actuator) Reconcile(ctx context.Context, _ logr.Logger, backupBucket *e
...
 	}
 
-	azCloudConfiguration, err := azureclient.AzureCloudConfigurationFromCloudConfiguration(backupConfig.CloudConfiguration)
+	cloudConfiguration := backupConfig.CloudConfiguration
+
+	if cloudConfiguration == nil {
+		cloudConfiguration = &azure.CloudConfiguration{Name: helper.CloudInstanceNameFromRegion(backupBucket.Spec.Region)}
+	}
+
+	azCloudConfiguration, err := azureclient.AzureCloudConfigurationFromCloudConfiguration(cloudConfiguration)
```
kon-angelo, line 50: Can't these functions be grouped up in a single function ? Can we imagine using one but not the other ?

AndreasBurger: Unfortunately, there is _one_ case where they are both required. However, I agree that duplicating these lines of code tenfold is not the way to go.

I've tried my hand at refactoring it a bit, but unfortunately there is some inherent ugliness here due to the way the upstream cloud-instance structs are defined. Not sure if I have hit the mark, honestly 🤔

kon-angelo (dismissed):
/lgtm

kon-angelo (approved):
/lgtm

PR #883, opened 2024-06-10
Fix possible Nullpointer-deref

AndreasBurger:

/area control-plane
/kind bug
/platform azure

Fixes a possible nullpointer-deref for zoned clusters that also give an external vnet config

Fixes # 

**Release note**:

```bugfix user
Fix panic that could occur when using zoned cluster and providing external vnet config
```


hebelsan (approved):
/lgtm

PR #931, opened 2024-08-01
update backupbucket fields

kon-angelo:

/area control-plane
/kind enhancement
/platform azure

Publish the `BackupBucket` domain name, to enable different Azure landscapes. 
See: https://github.com/gardener/etcd-backup-restore/pull/756

Fixes #

**Release note**:

```other operator
The BackupBucket controller now adds an additional field in the generated secret to indicate the blob storage service domain. This can be used to create blobs in other Azure environments like CN or USGov.
```


AndreasBurger (approved):

pkg/azure/client/storage.go
```diff
@@ -28,9 +30,46 @@ type BlobStorageClient struct {
...
-// newStorageClient creates a client for an Azure Blob storage by reading auth information from secret reference. Requires passing the storage domain (formerly
+// BlobStorageDomainFromCloudConfiguration returns the storage service domain given a known cloudConfiguration.
+func BlobStorageDomainFromCloudConfiguration(cloudConfiguration *azureapi.CloudConfiguration) (string, error) {
+	// Unfortunately the valid values for storage domains run by Microsoft do not seem to be part of any sdk module. They might be queryable from the cloud configuration,
+	// but I also haven't been able to find a documented list of proper ServiceName values.
+	// Furthermore, it seems there is still no unified way of specifying the cloud instance to connect to as the domain remains part of the storage account URL while
+	// the new options _also_ allow configuring the cloud instance.
+	switch {
+	case cloudConfiguration == nil:
+		return "blob.core.windows.net", nil
```
unmarshall, line 41: Is it possible to make constants for these? (It is a pity that designers of the SDK do not think of these things)
```go
const (
  BlobStoreGlobalDomain = "blob.core.windows.net"
  BlobStoreChinaDomain = "blob.core.chinacloudapi.cn"
  BlobStoreUSGovDomain = "blob.core.usgovcloudapi.net"
```

pkg/azure/client/storage.go
```diff
@@ -28,9 +30,46 @@ type BlobStorageClient struct {
...
+	case strings.EqualFold(cloudConfiguration.Name, "AzurePublic"):
```
unmarshall, line 42: you can club 2 cases into 1:
```go
case cloudConfiguration == nil:
case strings.EqualFold(cloudConfiguration.Name, "AzurePublic"):
  return "blob.core.windows.net", nil
```

pkg/azure/client/storage.go
```diff
@@ -28,9 +30,46 @@ type BlobStorageClient struct {
...
+	credentials, err := azblob.NewSharedKeyCredential(storageAccountName, storageAccountKey)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create shared key credentials: %v", err)
+	}
+
+	storageEndpointURL, err := url.Parse(fmt.Sprintf("https://%s.%s", storageAccountName, storageDomain))
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse service url: %v", err)
+	}
+	blobclient, err := azblob.NewClientWithSharedKeyCredential(storageEndpointURL.String(), credentials, nil)
```
unmarshall, line 66: If you look at the azure-sdk-for-go the nomenclature is a bit different. They provide 3 different clients AFAIK:

1. Blob client which is created with a specific blob URL (see [here](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/blob/client.go#L38))
2. Container client which is created for a specific container/bucket ( see [here](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/container/client.go#L44)).
3. Service client to manage buckets/containers (see [here](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/service/client.go#L42))

So what you have here is a `ContainerClient` or a `BucketClient` but you call it a blobClient. This confused me a bit.

unmarshall (commented):
Minor comments

pkg/azure/client/storage.go
renormalize: `fallthrough` must be used to achieve this functionality in Go, but there's a more idiomatic way of doing this.
```suggestion
	case cloutConfiguration == nil, strings.EqualFold(cloudConfiguration.Name, "AzurePublic"):
```

kon-angelo: I think that you would need to use `fallthrough` for that. Also in a very recent PR of mine in gardener core, they voted to merge conditions by `||`. I will probably do the same here for coherence

kon-angelo:
tested with http://europe-docker.pkg.dev/gardener-project/snapshots/gardener/etcd-druid:v0.22.4-dev-c692c9db187df47bc45c2d9815e5e31108b30faf successfully.

/unhold

AndreasBurger (approved):

PR #932, opened 2024-08-02
fix value passed to CCM

AndreasBurger:

/area control-plane
/kind bug
/platform azure

Fixes a bug introduced in the recent upgrade-PR that passes values to the CCM-chart in the wrong format.

**Release note**:

```other operator
NONE
```


hebelsan (approved):
/lgtm

PR #938, opened 2024-08-12
[Release v1.46] Default to Public Azure Instance in DNSRecord Actuator

AndreasBurger:

/area control-plane
/kind enhancement
/platform azure

The DNSRecord Actuator will now default to the Public Azure instance if no other information is available as (unlike for our other actuators) the required information is not guaranteed to be available.

Fixes #

**Release note**:

```other operator
DNSRecord Actuator client will now default to the public Azure cloud if no explicit config is given (instead of failing)
```


hebelsan (approved):
/lgtm

PR #945, opened 2024-08-20
Filter more strictly when looking at subnets

AndreasBurger:

/area usability
/kind regression
/platform azure

With this PR, subnet detection by name is more strict than before. This avoids erroneously detecting subnets that are deployed in the same vnet whose name only differs by a suffix from the current shoot when deploying into existing subnets.

**Release note**:

```other user
Flow reconciliation is now more strict when filtering subnets. This prevents subnets of shoots that are deployed into one Vnet from interfering with reconciliation.
```


hebelsan:
We still need to test this right? - I could take over

kon-angelo (approved):
/lgtm

PR #946, opened 2024-08-21
Infer cloud instance from bucket region when possible

AndreasBurger:

/area backup
/kind enhancement
/platform azure

The storage domain to use for backup buckets will now be inferred from the buckets' region if no explicit config is given. This is consistent with how the rest of our codebase does it.

**Release note**:

```other operator
The storage domain to use for backup buckets is now inferred from the buckets' region if no explicit config is given
```

### DIFF
--- a/pkg/controller/backupbucket/actuator.go
+++ b/pkg/controller/backupbucket/actuator.go
@@ -66,7 +66,12 @@ func (a *actuator) Reconcile(ctx context.Context, _ logr.Logger, backupBucket *e
 			return util.DetermineError(err, helper.KnownCodes)
 		}
 
-		storageDomain, err := azureclient.BlobStorageDomainFromCloudConfiguration(backupConfig.CloudConfiguration)
+		bucketCloudConfiguration, err := azureclient.CloudConfiguration(backupConfig.CloudConfiguration, &backupBucket.Spec.Region)
+		if err != nil {
+			return err
+		}
+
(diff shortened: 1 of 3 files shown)


pkg/controller/backupbucket/actuator.go
```diff
@@ -66,7 +66,12 @@ func (a *actuator) Reconcile(ctx context.Context, _ logr.Logger, backupBucket *e
 			return util.DetermineError(err, helper.KnownCodes)
 		}
 
-		storageDomain, err := azureclient.BlobStorageDomainFromCloudConfiguration(backupConfig.CloudConfiguration)
+		bucketCloudConfiguration, err := azureclient.CloudConfiguration(backupConfig.CloudConfiguration, &backupBucket.Spec.Region)
```
hebelsan, line 69: Can't we reuse `azCloudConfiguration` from line 45?

hebelsan (changes requested):
small comment

pkg/controller/backupbucket/actuator.go
AndreasBurger: No, that is an `cloud.Configuration` (azure SDK type) whereas this function requires a `azure.CloudConfiguration` (our type).

This is in part due to the way these types are defined in the sdk, which makes them rather unwieldy in general. Still, when interacting directly with the SDK clients (as in l.45, kind of) they are the way to go. But in other instances, our Type is much more reasonable to check against.

hebelsan (approved):

PR #947, opened 2024-08-21
[Release v1.47] Infer cloud instance from bucket region when possible

AndreasBurger:

/area backup
/kind enhancement
/platform azure

The storage domain to use for backup buckets will now be inferred from the buckets' region if no explicit config is given. This is consistent with how the rest of our codebase does it.

**Release note**:

```other operator
The storage domain to use for backup buckets is now inferred from the buckets' region if no explicit config is given
```


hebelsan (approved):
/lgtm

PR #975, opened 2024-09-30
Fix nil pointer dereference

hebelsan:

/area quality
/kind bug
/platform azure

Fix nil pointer dereference if diagnostic profile without given storage URI is used.

Fixes #

**Release note**:

```bugfix operator
Fix nil pointer dereference if diagnostic profile without given storage URI is used.
```


kon-angelo (approved):
/lgtm

PR #1025, opened 2024-11-29
Availability set deprecation

kon-angelo:

/area control-plane
/kind enhancement
/platform azure

- Deprecate Availability-Set based shoots
- Make VMSS-FLex (VMO) the default for non-zonal clusters
- Create a migration path based on https://learn.microsoft.com/en-us/azure/virtual-network/ip-services/public-ip-basic-upgrade-guidance to migrate AS to VMO shoots.

Fixes #

**Release note**:

```breaking user
Disable the creation of Availability-Set-based shoots.
```
```breaking user
VMSS-Flex based shoots are not the default deployment for non-zonal shoots.
```
```feature user
Introduce an annotation to migrate the availability-set shoots to VMSS-Flex shoots.
```

### DIFF
--- a/pkg/azure/client/loadbalancers.go
+++ b/pkg/azure/client/loadbalancers.go
@@ -25,6 +25,15 @@ func NewLoadBalancersClient(auth internal.ClientAuth, tc azcore.TokenCredential,
 	return &LoadBalancersClient{client}, err
 }
 
+// Get gets a given virtual load balancer by name
+func (c *LoadBalancersClient) Get(ctx context.Context, resourceGroupName, name string) (*armnetwork.LoadBalancer, error) {
+	res, err := c.client.Get(ctx, resourceGroupName, name, nil)
+	if err != nil {
+		return nil, FilterNotFoundError(err)
+	}
+	return &res.LoadBalancer, err
+}
+
 // List lists all subnets of a given virtual network.
 func (c *LoadBalancersClient) List(ctx context.Context, resourceGroupName string) ([]*armnetwork.LoadBalancer, error) {
 	pager := c.client.NewListPager(resourceGroupName, nil)
--- a/pkg/controller/infrastructure/infraflow/ensurer.go
+++ b/pkg/controller/infrastructure/infraflow/ensurer.go
@@ -178,7 +178,7 @@ func (fctx *FlowContext) EnsureAvailabilitySet(ctx context.Context) error {
 	}
 
 	// complete AS migration.
-	if v := fctx.whiteboard.GetChild("migration").GetChild(KindAvailabilitySet.String()).Get("complete"); v != nil && *v == "true" {
+	if v := fctx.whiteboard.GetChild(ChildKeyMigration).GetChild(KindAvailabilitySet.String()).Get(ChildKeyComplete); v != nil && *v == "true" {
 		err := fctx.deleteAvailabilitySet(ctx, log)
 		if err != nil {
 			return err
@@ -750,7 +750,7 @@ func (fctx *FlowContext) MigrateAvailabilitySet(ctx context.Context) error {
 	)
 
 	// return early if the migration has already been complete
-	if v := fctx.whiteboard.GetChild("migration").GetChild(KindAvailabilitySet.String()).Get("complete"); v != nil && *v == "true" {
+	if v := fctx.whiteboard.GetChild(ChildKeyMigration).GetChild(KindAvailabilitySet.String()).Get(ChildKeyComplete); v != nil && *v == "true" {
 		return nil
 	}
 	// return early if the cluster does not have AS.
@@ -820,7 +820,7 @@ func (fctx *FlowContext) MigrateAvailabilitySet(ctx context.Context) error {
 	if err := fctx.UpdatePublicIPs(ctx); err != nil {
 		return err
 	}
-	fctx.whiteboard.GetChild("migration").GetChild(KindAvailabilitySet.String()).Set("complete", "true")
+	fctx.whiteboard.GetChild(ChildKeyMigration).GetChild(KindAvailabilitySet.String()).Set(ChildKeyComplete, "true")
 	return fctx.PersistState(ctx)
 }
 
@@ -890,7 +890,7 @@ func (fctx *FlowContext) GetInfrastructureStatus(_ context.Context) (*v1alpha1.I
 				CountUpdateDomains: cfg.CountUpdateDomains,
 			},
 		}
-		if v := fctx.whiteboard.GetChild("migration").GetChild(KindAvailabilitySet.String()).Get("complete"); v != nil && *v == "true" {
+		if v := fctx.whiteboard.GetChild(ChildKeyMigration).GetChild(KindAvailabilitySet.String()).Get(ChildKeyComplete); v != nil && *v == "true" {
 			status.MigratingToVMO = true
 		}
 	}
--- a/pkg/controller/infrastructure/infraflow/ensurer.go
+++ b/pkg/controller/infrastructure/infraflow/ensurer.go
@@ -808,12 +808,15 @@ func (fctx *FlowContext) MigrateAvailabilitySet(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	log.Info("Deleting load balancer", "Name", fctx.infra.Namespace)
-	if err := loadbalancerClient.Delete(ctx, fctx.adapter.ResourceGroupName(), fctx.adapter.TechnicalName()); err != nil {
+	loadbalancerName := fctx.adapter.TechnicalName()
+	log.Info("Deleting load balancer", "Name", loadbalancerName)
+	if err := loadbalancerClient.Delete(ctx, fctx.adapter.ResourceGroupName(), loadbalancerName); err != nil {
 		return err
 	}
-	log.Info("Deleting internal load balancer", "Name", fctx.infra.Namespace)
-	if err := loadbalancerClient.Delete(ctx, fctx.adapter.ResourceGroupName(), fmt.Sprintf("%s-internal", fctx.adapter.TechnicalName())); err != nil {
+
+	loadbalancerName = fmt.Sprintf("%s-internal", fctx.adapter.TechnicalName())
+	log.Info("Deleting internal load balancer", "Name", loadbalancerName)
+	if err := loadbalancerClient.Delete(ctx, fctx.adapter.ResourceGroupName(), loadbalancerName); err != nil {
 		return err
 	}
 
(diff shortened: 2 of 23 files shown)


pkg/controller/infrastructure/infraflow/ensurer.go
```diff
@@ -699,6 +741,88 @@ func (fctx *FlowContext) EnsureManagedIdentity(ctx context.Context) (err error)
...
+		log.Info("Scaling deployment to 0 replicas", "Name", key.String())
+		deployment := &appsv1.Deployment{}
+		if err := fctx.client.Get(ctx, k8sclient.ObjectKey{
+			Namespace: fctx.infra.Namespace,
+			Name:      azure.CloudControllerManagerName,
+		}, deployment); k8sclient.IgnoreNotFound(err) != nil {
+			return err
+		} else if k8sclient.IgnoreNotFound(err) != nil {
+			return nil
+		}
```
AndreasBurger, line 776:
```suggestion
		if err := fctx.client.Get(ctx, k8sclient.ObjectKey{
			Namespace: fctx.infra.Namespace,
			Name:      azure.CloudControllerManagerName,
		}, deployment); k8sclient.IgnoreNotFound(err) != nil {
			return err
		}
```
Not sure if this is the right one to keep

pkg/controller/infrastructure/infraflow/flow_context.go
```diff
@@ -28,6 +29,14 @@ const (
 	defaultLongTimeout = 4 * time.Minute
 )
 
+var setSeparator = sync.OnceFunc(func() {
+	shared.Separator = "|"
```
AndreasBurger, line 33: I'm confused, Why do you change the separator here?
Also, is setSeparator the set separator or does it set the separator?

pkg/azure/client/loadbalancers.go
```diff
@@ -25,6 +25,15 @@ func NewLoadBalancersClient(auth internal.ClientAuth, tc azcore.TokenCredential,
 	return &LoadBalancersClient{client}, err
 }
 
+// Get gets a given virtual load balancer by name
+func (c *LoadBalancersClient) Get(ctx context.Context, resourceGroupName, name string) (*armnetwork.LoadBalancer, error) {
+	res, err := c.client.Get(ctx, resourceGroupName, name, nil)
+	if err != nil {
+		return nil, FilterNotFoundError(err)
```
AndreasBurger, line 32: Haven't checked around, but is this how we usually do it? This get method will always shadow notFound, I think this should be done when calling this function (if the caller wants it)

pkg/controller/infrastructure/infraflow/flow_context.go
kon-angelo: The `setSeparator` is supposed to change the separator used in the state once. The separator is changed from "/" to "|" because some of the strings that we save do use "/" internally and when the state is restored the slashes interfere with really restoring the state.

We can refactor the strings that we save, or escape the strings or just change the separator. Previous to this PR (and I believe after removing much of the code post-deprecation) we didn't really need to restore things from the state which is why this is only introduced now.

pkg/azure/client/loadbalancers.go
kon-angelo: This is also how the other client implementations work. The comment is fair, but for local consistency it would be applied in a separate PR.

pkg/controller/infrastructure/infraflow/migrations.go
```diff
@@ -0,0 +1,97 @@
...
+		lb.SKU == nil ||
+		lb.SKU.Name == nil ||
+		*lb.SKU.Name != armnetwork.LoadBalancerSKUNameBasic ||
+		lb.Properties == nil ||
+		len(lb.Properties.FrontendIPConfigurations) == 0 {
+		return nil
+	}
+	if lb.Properties == nil || len(lb.Properties.FrontendIPConfigurations) == 0 {
+		return nil
+	}
```
hebelsan, line 40: duplicated check

pkg/controller/infrastructure/infraflow/ensurer.go
```diff
@@ -699,6 +742,88 @@ func (fctx *FlowContext) EnsureManagedIdentity(ctx context.Context) (err error)
...
+	}
+	// return early if the cluster does not have AS.
+	if fctx.whiteboard.GetChild(ChildKeyIDs).Get(KindAvailabilitySet.String()) == nil {
+		return nil
+	}
+
+	// IF VMOs are not needed, or the migration is already done, return early.
+	if !helper.HasShootVmoMigrationAnnotation(fctx.cluster.Shoot.GetAnnotations()) {
+		return nil
+	}
```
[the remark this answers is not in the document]
kon-angelo: These checks as a group are used in one place. The separate function will just be giving a name to preflight checks but I can oblige

pkg/controller/infrastructure/infraflow/infra_adapter.go
```diff
@@ -157,28 +157,41 @@ type AvailabilitySetConfig struct {
...
+	if len(ia.status.AvailabilitySets) > 0 {
+		if _, err := helper.FindAvailabilitySetByPurpose(ia.status.AvailabilitySets, azure.PurposeNodes); err == nil {
+			return true
+		}
+	}
+	return false
+}
+
+// IsVmoRequiredForInfrastructure determines if VMO is required.
+func (ia *InfrastructureAdapter) IsVmoRequiredForInfrastructure() bool {
```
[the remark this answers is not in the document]
kon-angelo: old artifact - thank you

pkg/controller/infrastructure/infraflow/ensurer.go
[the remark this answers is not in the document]
kon-angelo: - Do you mean as a `shared.DoIf` ?
It's a question of when are the dependencies of a task evaluated (at build or at runtime) and AFAIK they are on build time. At that point you may not have the `KindAvailabilitySet` in the state for example. 

- Do you mean as a `shared.Dependency` ? 
I am actually not sure if this will block a child Task in the graph from being executed. I would need to test the scemantics. It may either mark the skipped parent as completed (so the childen are always executed), or it may really prune the tree (what you suggested I believe). But I need to check first.

pkg/controller/infrastructure/infraflow/flow_context.go
kon-angelo: > Also, is setSeparator the set separator or does it set the separator?

Is the function to set the separator once. It is executed on package import with the `init` couple lines below.

pkg/controller/infrastructure/infraflow/ensurer.go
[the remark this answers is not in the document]
kon-angelo: Based on https://github.com/gardener/gardener/blob/84184385dbb9616c2c838e4f24487cae2be75da0/pkg/utils/flow/flow.go#L303 I think it marks the skipped nodes as completed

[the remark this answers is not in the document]
hebelsan: Valid arguments.
I think I was a bit too quick here and didn't think it through completely.
Let's leave it as is.

pkg/controller/infrastructure/infraflow/ensurer.go
```diff
@@ -699,6 +742,88 @@ func (fctx *FlowContext) EnsureManagedIdentity(ctx context.Context) (err error)
...
+		return err
+	}
+	log.Info("Deleting load balancer", "Name", fctx.infra.Namespace)
+	if err := loadbalancerClient.Delete(ctx, fctx.adapter.ResourceGroupName(), fctx.adapter.TechnicalName()); err != nil {
+		return err
+	}
+	log.Info("Deleting internal load balancer", "Name", fctx.infra.Namespace)
+	if err := loadbalancerClient.Delete(ctx, fctx.adapter.ResourceGroupName(), fmt.Sprintf("%s-internal", fctx.adapter.TechnicalName())); err != nil {
+		return err
+	}
```
AndreasBurger, line 818:
```suggestion
	}
	loadbalancerName := fctx.adapter.TechnicalName()
	log.Info("Deleting load balancer", "Name", loadbalancerName)
	if err := loadbalancerClient.Delete(ctx, fctx.adapter.ResourceGroupName(), loadbalancerName); err != nil {
		return err
	}
	loadbalancerName = fmt.Sprintf("%s-internal", fctx.adapter.TechnicalName())
	log.Info("Deleting internal load balancer", "Name", loadbalancerName )
	if err := loadbalancerClient.Delete(ctx, fctx.adapter.ResourceGroupName(), loadbalancerName ); err != nil {
		return err
	}
```

pkg/controller/infrastructure/infraflow/ensurer.go
```diff
@@ -699,6 +742,88 @@ func (fctx *FlowContext) EnsureManagedIdentity(ctx context.Context) (err error)
...
+		}
+
+		if ptr.Deref(deployment.Spec.Replicas, 1) == 0 {
+			return nil
+		}
+		patch := k8sclient.MergeFrom(deployment.DeepCopy())
+		deployment.Spec.Replicas = ptr.To(int32(0))
+		// Apply the patch on the "scale" subresource
+		if err := c.SubResource("scale").Patch(ctx, deployment, patch); err != nil {
+			return fmt.Errorf("failed to patch deployment %s with replicas: %w", key.String(), err)
```
AndreasBurger, line 786:
```suggestion
			return fmt.Errorf("failed to patch deployment %s with replicas: %w", key, err)
```
should not be needed

pkg/controller/infrastructure/infraflow/infra_adapter.go
```diff
@@ -157,28 +157,36 @@ type AvailabilitySetConfig struct {
...
 
-// AvailabilitySetRequired returns true if gardener should create an availability set for the shoot.
-func (ia *InfrastructureAdapter) availabilitySetRequired() (bool, error) {
-	return infrastructure.IsPrimaryAvailabilitySetRequired(ia.infra, ia.config, ia.cluster)
+// IsAvailabilitySetReconciliationRequired returns true if gardener should create an availability set for the shoot.
+func (ia *InfrastructureAdapter) IsAvailabilitySetReconciliationRequired() bool {
+	if ia.config.Zoned {
+		return false
+	}
+	// If the infrastructureStatus already exists that mean the Infrastucture is already created.
```
AndreasBurger, line 165:
```suggestion
	// If the infrastructureStatus already exists that means the Infrastucture is already created.
```

AndreasBurger (dismissed):

AndreasBurger (approved):

PR #1037, opened 2024-12-10
Fix flow reconcile ensure public ip  release v.1.49

hebelsan:

/area usability
/kind bug
/platform azure

If a public IP was created by the azure CCM (Cloud Controller Manager) in the past with a pretty old CCM version it was tagged under the key "service" and not "k8s-azure-service".
The flow code up until now does not support this old key and therefore the reconciliation fails.

Fixes #

**Release note**:

```bugfix user
Support legacy CCM service tag key in flow reconciliation
```


AndreasBurger (approved):
lgtm

PR #1051, opened 2024-12-27
Fix reconciliation of user managed public IPs in flow

hebelsan:

/area quality
/kind bug
/platform azure

This PR prevents the deletion of certain user managed public IPs during flow reconciliation after Terraform migration.
It adds a tag to IPs that reside in the shoot namespace and have the shoot prefix but should not be touched by us.

Fixes #

**Release note**:

```bugfix user
This PR prevents the deletion of certain user managed public IPs during flow reconciliation after Terraform migration
```


hebelsan:
Close in favour of https://github.com/gardener/gardener-extension-provider-azure/pull/1052

PR #1052, opened 2024-12-27
Fix reconciliation of user managed public IPs in flow

hebelsan:

/area quality
/kind bug
/platform azure

This PR prevents the deletion of certain user managed public IPs during flow reconciliation after Terraform migration.
It adds a tag to IPs that reside in the shoot namespace and have the shoot prefix but should not be touched by us.

Fixes #

**Release note**:

```bugfix user
This PR prevents the deletion of certain user managed public IPs during flow reconciliation after Terraform migration
```

### DIFF
--- a/pkg/controller/infrastructure/infraflow/ensurer.go
+++ b/pkg/controller/infrastructure/infraflow/ensurer.go
@@ -398,7 +398,10 @@ func (fctx *FlowContext) ensurePublicIps(ctx context.Context) error {
 	}
 	currentIPs = Filter(currentIPs, func(address *armnetwork.PublicIPAddress) bool {
 		// filter only these IpConfigs that are managed by gardener
-		return fctx.adapter.HasShootPrefix(address.Name) && address.Tags[ManagedByGardenerTag] != nil
+		return fctx.adapter.HasShootPrefix(address.Name) &&
+			ptr.Deref(address.Tags[TagManagedByGardener], "") == "true" &&
+			ptr.Deref(address.Tags[TagShootName], "") == fctx.adapter.TechnicalName()
+
 	})
 	// obtain an indexed list of current IPs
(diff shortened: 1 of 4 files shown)


pkg/controller/infrastructure/infraflow/infra_adapter.go
```diff
@@ -540,6 +541,11 @@ func (ip *PublicIPConfig) ToProvider(base *armnetwork.PublicIPAddress) *armnetwo
 		target.Tags = base.Tags
 	}
 
+	// TODO(hebelsan) remove in later release because we add the key when creating the object.
+	if _, ok := target.Tags[ManagedByGardenerTag]; !ok {
+		target.Tags[ManagedByGardenerTag] = to.Ptr("true")
+	}
+
```
AndreasBurger, line 548: I'd prefer a merge of `target.Tags` into `base.Tags` in l.541.

hebelsan:
Tested migration from terraform to flow with customer managed public IP and shoot prefix name.
The IP is no longer deleted by the flow.

hebelsan:
Successfully tested that the public IP stays the same after switching to this controller version.

pkg/controller/infrastructure/infraflow/ensurer.go
```diff
@@ -424,6 +427,13 @@ func (fctx *FlowContext) ensurePublicIps(ctx context.Context) error {
 		// delete all the resources that are not in the list of target resources
 		pipCfg, ok := desiredConfiguration[name]
 		if !ok {
+			// 	skip deletion of IPs that may fit our criteria, but are referenced explicitly in the shoot spec.
```
hebelsan, line 430:
```suggestion
			// skip deletion of IPs that may fit our criteria, but are referenced explicitly in the shoot spec.
```

hebelsan (commented):
just a few formatting notes

kon-angelo (approved):
/lgtm

Issue #1056, opened 2025-01-10
Several `nil pointer dereference` panics


ialidzhikov:
**How to categorize this issue?**

/area quality
/kind bug
/platform azure

**What happened**:
With @dnaeon we noticed the following provider-azure panic:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x235aae6]

goroutine 3177 [running]:
github.com/gardener/gardener-extension-provider-azure/pkg/controller/infrastructure/infraflow.(*FlowContext).ensureSubnets(0xc003d46960, {0x2fd4270, 0xc000152a10})
	/go/src/github.com/gardener/gardener-extension-provider-azure/pkg/controller/infrastructure/infraflow/ensurer.go:595 +0x366
github.com/gardener/gardener-extension-provider-azure/pkg/controller/infrastructure/infraflow.(*FlowContext).EnsureSubnets(...)
	/go/src/github.com/gardener/gardener-extension-provider-azure/pkg/controller/infrastructure/infraflow/ensurer.go:561
github.com/gardener/gardener-extension-provider-azure/pkg/controller/infrastructure/infraflow/shared.(*BasicFlowContext).AddTask.TaskFn.Timeout.func1({0x2fd41c8?, 0xc00a84a330?})
	/go/pkg/mod/github.com/gardener/gardener@v1.108.0/pkg/utils/flow/taskfn.go:35 +0x59
github.com/gardener/gardener-extension-provider-azure/pkg/controller/infrastructure/infraflow/shared.(*BasicFlowContext).AddTask.(*BasicFlowContext).wrapTaskFn.func2({0x2fd41c8, 0xc00a84a330})
	/go/src/github.com/gardener/gardener-extension-provider-azure/pkg/controller/infrastructure/infraflow/shared/basic_context.go:169 +0x390
github.com/gardener/gardener/pkg/utils/flow.(*execution).runNode.func2()
	/go/pkg/mod/github.com/gardener/gardener@v1.108.0/pkg/utils/flow/flow.go:226 +0x165
created by github.com/gardener/gardener/pkg/utils/flow.(*execution).runNode in goroutine 759
	/go/pkg/mod/github.com/gardener/gardener@v1.108.0/pkg/utils/flow/flow.go:222 +0x4f0
```

[...]

PR #1057, opened 2025-01-10
fix inventory for deleted resources

kon-angelo:

/area control-plane
/kind bug
/platform azure

Fixes https://github.com/gardener/gardener-extension-provider-azure/issues/1056

**Release note**:

```bugfix operator
Fix an issue causing nil pointer exception when the remote resource that was once existing and present in the inventory, had been deleted during reconciliation
```

### DIFF
--- a/pkg/controller/infrastructure/infraflow/ensurer.go
+++ b/pkg/controller/infrastructure/infraflow/ensurer.go
@@ -318,7 +318,7 @@ func (fctx *FlowContext) EnsureSecurityGroup(ctx context.Context) error {
 		return err
 	}
 
-	log.V(1).Info("adding to inventory", "id", *sg.ID)
+	log.V(1).Info("Adding to inventory", "id", *sg.ID)
 	err = fctx.inventory.Insert(*sg.ID)
 	if err != nil {
 		return err
@@ -426,10 +426,10 @@ func (fctx *FlowContext) ensurePublicIps(ctx context.Context) error {
 	for name, ip := range desiredConfiguration {
 		toReconcile[name] = ip.ToProvider(nameToCurrentIps[name])
 	}
-	for _, inv := range fctx.inventory.ByKind(KindPublicIP) {
-		if _, ok := nameToCurrentIps[inv.Name]; !ok {
-			log.V(1).Info("removing public IP from inventory", "id", inv.String())
-			fctx.inventory.Delete(inv.String())
+	for _, resource := range fctx.inventory.ByKind(KindPublicIP) {
+		if _, ok := nameToCurrentIps[resource.Name]; !ok {
+			log.Info("Removing public IP from inventory", "id", resource)
+			fctx.inventory.Delete(resource.String())
 		}
 	}
 
@@ -440,14 +440,14 @@ func (fctx *FlowContext) ensurePublicIps(ctx context.Context) error {
 		// delete all the resources that are not in the list of target resources
 		pipCfg, ok := desiredConfiguration[name]
 		if !ok {
-			log.Info("will delete public IP because it is not needed", "Resource Group", fctx.adapter.ResourceGroupName(), "Name", name)
+			log.Info("Will delete public IP because it is not needed", "Resource Group", fctx.adapter.ResourceGroupName(), "Name", name)
 			toDelete[name] = *current.ID
 			continue
 		}
 
 		// delete all resources whose spec cannot be updated to match target spec.
 		if ok, offender, v := ForceNewIp(current, toReconcile[pipCfg.Name]); ok {
-			log.Info("will delete public IP because it can't be reconciled", "Resource Group", fctx.adapter.ResourceGroupName(), "Name", name, "Field", offender, "Value", v)
+			log.Info("Will delete public IP because it can't be reconciled", "Resource Group", fctx.adapter.ResourceGroupName(), "Name", name, "Field", offender, "Value", v)
 			toDelete[name] = *current.ID
 			continue
 		}
@@ -524,10 +524,10 @@ func (fctx *FlowContext) ensureNatGateways(ctx context.Context) error {
 		toReconcile[name] = target
 	}
 
-	for _, inv := range fctx.inventory.ByKind(KindNatGateway) {
-		if _, ok := nameToCurrentNats[inv.Name]; !ok {
-			log.V(1).Info("removing nat gateway from inventory", "id", inv.String())
-			fctx.inventory.Delete(inv.String())
+	for _, resource := range fctx.inventory.ByKind(KindNatGateway) {
+		if _, ok := nameToCurrentNats[resource.Name]; !ok {
+			log.Info("Removing nat gateway from inventory", "id", resource)
+			fctx.inventory.Delete(resource.String())
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/controller/infrastructure/infraflow/ensurer.go
```diff
@@ -632,9 +634,10 @@ func (fctx *FlowContext) ensureSubnets(ctx context.Context) (err error) {
 		return *s.Name
 	})
 	// clean the current inventory and rebuild it.
-	for _, name := range fctx.inventory.ByKind(KindSubnet) {
-		if subnet, ok := mappedSubnets[name]; !ok {
-			fctx.inventory.Delete(*subnet.ID)
+	for _, resource := range fctx.inventory.ByKind(KindSubnet) {
+		if _, ok := mappedSubnets[resource.Name]; !ok {
+			log.V(1).Info("removing subnet from inventory", "id", resource.String())
```
ialidzhikov, line 639: According to the [Gardener's logging guide](https://github.com/gardener/gardener/blob/v1.110.2/docs/development/logging.md) for developers and [Kubernetes' logging guide for developers](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md), log messages should start from a capital letter.

```suggestion
			log.V(1).Info("Removing subnet from inventory", "id", resource.String())
```

You may want to adapt all occurrences across the code base in another PR.

ialidzhikov, line 639: No need to explicitly invoke `.String()`, the library should do it already.

```suggestion
			log.V(1).Info("Removing subnet from inventory", "id", resource)
```

ialidzhikov, line 639: Why it is a debug log? IMO it deserves to be a log shown by default because it won't be shown always as some other log entries that will be displayed always on reconciliation like "adding resource to inventory". During investigations this log won't be shown and the log level should be explicitly raised on the extension side via flag.

```suggestion
			log.Info("Removing subnet from inventory", "id", resource)
```

pkg/controller/infrastructure/infraflow/ensurer.go
```diff
@@ -524,8 +525,9 @@ func (fctx *FlowContext) ensureNatGateways(ctx context.Context) error {
 	}
 
 	for _, inv := range fctx.inventory.ByKind(KindNatGateway) {
-		if nat, ok := nameToCurrentNats[inv]; !ok {
-			fctx.inventory.Delete(*nat.ID)
+		if _, ok := nameToCurrentNats[inv.Name]; !ok {
+			log.V(1).Info("removing nat gateway from inventory", "id", inv.String())
```
ialidzhikov, line 529:
```suggestion
			log.Info("Removing nat gateway from inventory", "id", resource)
```

pkg/controller/infrastructure/infraflow/ensurer.go
```diff
@@ -427,8 +427,9 @@ func (fctx *FlowContext) ensurePublicIps(ctx context.Context) error {
 		toReconcile[name] = ip.ToProvider(nameToCurrentIps[name])
 	}
 	for _, inv := range fctx.inventory.ByKind(KindPublicIP) {
-		if ip, ok := nameToCurrentIps[inv]; !ok {
-			fctx.inventory.Delete(*ip.ID)
+		if _, ok := nameToCurrentIps[inv.Name]; !ok {
+			log.V(1).Info("removing public IP from inventory", "id", inv.String())
```
ialidzhikov, line 431:
```suggestion
			log.Info("Removing public IP from inventory", "id", resource)
```

pkg/controller/infrastructure/infraflow/ensurer.go
```diff
@@ -524,8 +525,9 @@ func (fctx *FlowContext) ensureNatGateways(ctx context.Context) error {
 	}
 
 	for _, inv := range fctx.inventory.ByKind(KindNatGateway) {
```
ialidzhikov, line 527:
```suggestion
	for _, resource := range fctx.inventory.ByKind(KindNatGateway) {
```

pkg/controller/infrastructure/infraflow/ensurer.go
```diff
@@ -427,8 +427,9 @@ func (fctx *FlowContext) ensurePublicIps(ctx context.Context) error {
 		toReconcile[name] = ip.ToProvider(nameToCurrentIps[name])
 	}
 	for _, inv := range fctx.inventory.ByKind(KindPublicIP) {
```
ialidzhikov, line 429:
```suggestion
	for _, resource := range fctx.inventory.ByKind(KindPublicIP) {
```

pkg/controller/infrastructure/infraflow/ensurer.go
```diff
@@ -318,7 +318,7 @@ func (fctx *FlowContext) EnsureSecurityGroup(ctx context.Context) error {
 		return err
 	}
 
-	log.V(1).Info("adding to inventory", *sg.ID)
+	log.V(1).Info("adding to inventory", "id", *sg.ID)
```
ialidzhikov, line 321:
```suggestion
	log.V(1).Info("Adding to inventory", "id", *sg.ID)
```

ialidzhikov (commented):
Thanks for the fix! Some cosmetic suggestions inline.

pkg/controller/infrastructure/infraflow/ensurer.go
```diff
@@ -426,9 +426,10 @@ func (fctx *FlowContext) ensurePublicIps(ctx context.Context) error {
 	for name, ip := range desiredConfiguration {
 		toReconcile[name] = ip.ToProvider(nameToCurrentIps[name])
 	}
-	for _, inv := range fctx.inventory.ByKind(KindPublicIP) {
-		if ip, ok := nameToCurrentIps[inv]; !ok {
-			fctx.inventory.Delete(*ip.ID)
+	for _, resource := range fctx.inventory.ByKind(KindPublicIP) {
+		if _, ok := nameToCurrentIps[resource.Name]; !ok {
+			log.Info("Removing public IP from inventory", "id", resource)
+			fctx.inventory.Delete(resource.String())
```
AndreasBurger, line 432:
```suggestion
	for _, id := range fctx.inventory.ByKind(KindPublicIP) {
		if _, ok := nameToCurrentIps[id.Name]; !ok {
			log.Info("Removing public IP from inventory", "id", id.String())
			fctx.inventory.Delete(id.String())
```
Also, looking at the upstream code, `.Name` and `.String()` are not necessarily equal. It's a bit confusing that we have lookups for both, but I guess this is just an adjustment to the new `ByKind()`.

The renaming remark also applies to the other `for _, resource` loops, of course.

AndreasBurger (approved):

pkg/controller/infrastructure/infraflow/ensurer.go
kon-angelo: In the end, I decided to keep the `.String()`. The `String()` implementation here contains quite some logic to rebuild the actual ID string from the struct. and without the function call it prints the struct similar to a `%v` call.

AndreasBurger (approved):

pkg/controller/infrastructure/infraflow/ensurer.go
ialidzhikov: Ok, it makes sense to keep `.String()` in that case.

PR #1076, opened 2025-01-29
feat: Support for Kubernetes v1.32

marc1404:

/area open-source
/kind enhancement
/platform azure

This PR adds support for Kubernetes 1.32 to the extension.

Part of https://github.com/gardener/gardener/issues/11020

The previous PR for [supporting Kubernetes v1.31](https://github.com/gardener/gardener-extension-provider-azure/pull/963) has been used as a guide.

The functionality has been validated as follows:
* [x] Create new clusters with versions < 1.32
* [x] Create new clusters with version  = 1.32
* [x] Upgrade old clusters from version 1.31 to version 1.32
* [x] Delete clusters with versions < 1.32
* [x] Delete clusters with version  = 1.32

**Release note**:

```feature user
The provider-azure extension does now support shoot clusters with Kubernetes version 1.32. You should consider the [Kubernetes release notes](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.32.md) before upgrading to 1.32. 
```

### DIFF
--- a/charts/internal/shoot-system-components/charts/cloud-controller-manager/templates/rbac-cloud-controller.yaml
+++ b/charts/internal/shoot-system-components/charts/cloud-controller-manager/templates/rbac-cloud-controller.yaml
@@ -112,6 +112,7 @@ metadata:
   name: system:azure-cloud-provider
   annotations:
     rbac.authorization.kubernetes.io/autoupdate: "true"
+    resources.gardener.cloud/keep-object: "true"
   labels:
     k8s-app: cloud-controller-manager
 rules:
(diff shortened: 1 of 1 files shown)


imagevector/images.yaml
```diff
@@ -117,6 +118,21 @@ images:
       integrity_requirement: 'high'
       availability_requirement: 'low'
```
LucaBernstein, line 120: Please fix `targetVersion` selector in previous entry: `targetVersion: ">= 1.31"`

imagevector/images.yaml
```diff
@@ -222,6 +238,21 @@ images:
       integrity_requirement: 'high'
```
LucaBernstein, line 238: same as above.

LucaBernstein (changes requested):

imagevector/images.yaml
marc1404: My bad, thanks! Fixed 🫡

imagevector/images.yaml
marc1404: My bad, thanks! Fixed 🫡

LucaBernstein:
/hold
investigating an issue, where a service does not get an external ip due to seemingly service account permission issues.

AndreasBurger (approved):

PR #1098, opened 2025-02-24
Add support for container-level immutability to Azure Blob Storage Containers

renormalize:

/area backup security
/kind enhancement
/platform azure

Mirror of https://github.com/gardener/gardener-extension-provider-gcp/pull/906

Part of https://github.com/gardener/gardener/issues/10866

Tests performed:

| Before           | Before's Lock state | After            | After's Lock state | Test                                                                            | Output   | Expectation |
| ---------------- | ------------------- | ---------------- | ------------------ | ------------------------------------------------------------------------------- | -------- | ----------- |
| No immutability  | Unlocked            | 23h immutability | Unlocked           | Tests < 24h policy create rejection                                             | Rejected | Rejected    |
| No immutability  | Unlocked            | 24h immutability | Unlocked           | Tests policy creation                                                           | Accepted | Accepted    |
| No immutability  | Unlocked            | 25h immutability | Unlocked           | Tests non-multiple of 24h policy create rejection                               | Rejected | Rejected    |
| 24h immutability | Unlocked            | 23h immutability | Unlocked           | Tests < 24h policy update rejection                                             | Rejected | Rejected    |
| 24h immutability | Unlocked            | 25h immutability | Unlocked           | Tests non-multiple of 24h policy  update rejection                              | Rejected | Rejected    |
| 24h immutability | Unlocked            | 48h immutability | Unlocked           | Tests policy update while unlocked                                              | Accepted | Accepted    |
| 48h immutability | Unlocked            | 24h immutability | Unlocked           | Tests policy reduction while unlocked                                           | Accepted | Accepted    |
| 48h immutability | Unlocked            | No immutability  | Unlocked           | Set `immutability` to `null`<br><br>Tests policy deletion while unlocked        | Accepted | Accepted    |
| 48h immutability | Unlocked            | No immutability  | Unlocked           | Remove the `immutability` section<br><br>Tests policy deletion while unlocked   | Accepted | Accepted    |
| 48h immutability | Unlocked            | No immutability  | Unlocked           | Remove the `providerConfig` section<br><br>Tests policy deletion while unlocked | Accepted | Accepted    |
| 24h immutability | Unlocked            | 24h immutability | Locked             | Tests locking the policy                                                        | Accepted | Accepted    |
| 24h immutability | Locked              | 23h immutability | Locked             | Tests reduction to an invalid value after locking                               | Rejected | Rejected    |
| 24h immutability | Locked              | 25h immutability | Locked             | Tests extension to an invalid value after locking                               | Rejected | Rejected    |
| 24h immutability | Locked              | 24h immutability | Unlocked           | Tests removal of lock on the policy                                             | Rejected | Rejected    |
| 24h immutability | Locked              | No immutability  | -                  | Tests deletion of the immutability policy on a locked policy                    | Rejected | Rejected    |
| 24h immutability | Locked              | 48h immutability | Locked             | Tests policy extension while locked                                             | Accepted | Accepted    |
| 48h immutability | Locked              | 24h immutability | Locked             | Tests reduction to a valid value after locking                                  | Rejected | Rejected    |

This PR introduces a new client `BlobContainersClient` which is a client for ABS containers.
The reason this new client is introduced is because the storage client that was previously used did not expose methods for making the necessary API calls to Azure to modify buckets to become container-level immutable.
These changes also make the container client more uniform with the rest of the codebase.

**Release note**:

```noteworthy operator
Azure Blob Storage Containers can now be configured to be created with container-level immutability settings. 
```

```feature operator
The immutable bucket feature is currently in alpha state. The immutable bucket feature can be enabled by enabling `config.featureGates.enableImmutableBuckets`. More information can be found in `docs/usage/usage.md`.
```

```breaking operator
Enabling the immutable bucket feature is a breaking change, as it requires additional permissions. The permissions given to your credential are to be first modified before enabling the alpha feature.
```


renormalize:
Rebasing since there seems to be merge conflicts with master.

pkg/admission/validator/seed.go
```diff
@@ -0,0 +1,149 @@
...
+	if newSeed != nil && newSeed.Spec.Backup != nil && newSeed.Spec.Backup.ProviderConfig != nil {
+		newBackupBucketConfig, err = helper.BackupConfigFromProviderConfig(newSeed.Spec.Backup.ProviderConfig)
+		if err != nil {
+			allErrs = append(allErrs, field.Invalid(providerConfigfldPath, newSeed.Spec.Backup.ProviderConfig, fmt.Errorf("failed to decode new provider config: %v", err).Error()))
+			return allErrs
+		}
+	}
+
+	allErrs = append(allErrs, azurevalidation.ValidateBackupBucketConfig(&newBackupBucketConfig, providerConfigfldPath)...)
+	allErrs = append(allErrs, s.validateImmutabilityUpdate(&oldBackupBucketConfig, &newBackupBucketConfig, providerConfigfldPath)...)
```
kon-angelo, line 115: I would maybe prefer to move this checks with the other checks in `pkg/apis/validation`. I think this split of logic when checking will cause more of a headache since half due to the file splitting. WDYT?

pkg/controller/worker/machines.go
```diff
@@ -329,7 +329,7 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 		}
 
 		// AvailabilitySet
-		if !infrastructureStatus.Zoned && vmoDependency == nil {
+		if !infrastructureStatus.Zoned {
```
kon-angelo, line 332: Why this change ?

pkg/controller/backupbucket/actuator.go
```diff
@@ -58,34 +58,130 @@ func (a *actuator) Reconcile(ctx context.Context, _ logr.Logger, backupBucket *e
...
+	ctx context.Context, logger logr.Logger,
+	blobContainersClient azureclient.BlobContainers,
+	backupBucketConfig azure.BackupBucketConfig,
+	resourceGroupName, storageAccountName, backupBucketName string,
+	currentContainerImmutabilityDays *int32, currentlyLocked bool, etag *string,
+) error {
+	var currentDays, desiredDays int32
+	if currentContainerImmutabilityDays != nil {
+		currentDays = *currentContainerImmutabilityDays
+	}
```
kon-angelo, line 159:
```
currentDays = ptr.Deref(currentContainerImmutabilityDays, 0)
```
should work right ?

pkg/controller/backupbucket/actuator.go
```diff
@@ -58,34 +58,130 @@ func (a *actuator) Reconcile(ctx context.Context, _ logr.Logger, backupBucket *e
...
+
+	// lock the policy if configured
+	if backupBucketConfig.Immutability != nil && backupBucketConfig.Immutability.Locked && !currentlyLocked {
+		logger.Info("Locking bucket immutability policy")
+		// Etag of the resource will change if the policy was updated in the same reconiliation
+		// get the bucket policy again and use that etag
+		_, _, etag, err = blobContainersClient.GetImmutabilityPolicy(ctx, resourceGroupName, storageAccountName, backupBucket.Name)
+		if err != nil {
+			logger.Error(err, "Errored while fetching immutability information again before locking", "bucket", backupBucket.Name)
+			return util.DetermineError(err, helper.KnownCodes)
```
kon-angelo, line 137: Do you need or do any special handling for 404 already ?

pkg/controller/backupbucket/actuator.go
```diff
@@ -58,34 +58,130 @@ func (a *actuator) Reconcile(ctx context.Context, _ logr.Logger, backupBucket *e
...
+			return err
+		}
+	}
+
+	// update the bucket if necessary
+	currentContainerImmutabilityDays, currentlyLocked, etag, err := blobContainersClient.GetImmutabilityPolicy(ctx, resourceGroupName, storageAccountName, backupBucket.Name)
+	if err != nil {
+		logger.Error(err, "Errored while fetching immutability information", "bucket", backupBucket.Name)
+		return util.DetermineError(err, helper.KnownCodes)
+	}
```
kon-angelo, line 118: Same here for the 404

pkg/controller/backupbucket/actuator.go
```diff
@@ -58,34 +58,130 @@ func (a *actuator) Reconcile(ctx context.Context, _ logr.Logger, backupBucket *e
...
+
+	// Delete the policy if requested
+	if currentDays != 0 && desiredDays == 0 {
+		logger.Info("Deleting the bucket immutability policy")
+		return blobContainersClient.DeleteImmutabilityPolicy(ctx, resourceGroupName, storageAccountName, backupBucketName, etag)
+	}
+
+	// Update the policy if requested
+	if currentDays != desiredDays {
```
kon-angelo, line 180: Can you help me understand when we reach this part logically ? Is this just if we decided to change the number of days we want the policy locked?

pkg/controller/backupbucket/actuator.go
```diff
@@ -58,34 +58,130 @@ func (a *actuator) Reconcile(ctx context.Context, _ logr.Logger, backupBucket *e
...
+		currentContainerImmutabilityDays, currentlyLocked, etag,
+	); err != nil {
+		logger.Error(err, "Errored while updating the bucket")
+		return util.DetermineError(err, helper.KnownCodes)
+	}
+
+	// lock the policy if configured
+	if backupBucketConfig.Immutability != nil && backupBucketConfig.Immutability.Locked && !currentlyLocked {
+		logger.Info("Locking bucket immutability policy")
+		// Etag of the resource will change if the policy was updated in the same reconiliation
```
kon-angelo, line 132: typo: `reconciliation`

pkg/controller/backupbucket/actuator.go
```diff
@@ -58,34 +58,130 @@ func (a *actuator) Reconcile(ctx context.Context, _ logr.Logger, backupBucket *e
...
+		resourceGroupName, storageAccountName, backupBucket.Name,
+		currentContainerImmutabilityDays, currentlyLocked, etag,
+	); err != nil {
+		logger.Error(err, "Errored while updating the bucket")
+		return util.DetermineError(err, helper.KnownCodes)
+	}
+
+	// lock the policy if configured
+	if backupBucketConfig.Immutability != nil && backupBucketConfig.Immutability.Locked && !currentlyLocked {
+		logger.Info("Locking bucket immutability policy")
```
kon-angelo, line 131: Nit: but I can see this section also going in the `updateBackupBucketIfNeeded`. Maybe it's my preference but it's nice to know that all the logic for the immutability at least resides in one place. You could also technically track if you did any operation that would require you to do the second GET for the etag, or even have the `updateBackup...` return an etag that you can check.

Though I do not think that these optimisations are really necessary, it's just an idea for improvement

kon-angelo (changes requested):

pkg/admission/validator/seed.go
renormalize: That makes sense, was just trying to keep consistency with the GCP PR, so had the validations split across a similar way.

Will make the necessary changes.

pkg/controller/worker/machines.go
renormalize: It was just a tautology that was causing my LSP to crash somehow. The `&& vmoDependency == nil` check is actually unnecessary in the line since a few lines above, `vmoDependency != nil` is checked for, and if it is non-`nil`, the loop is `continue`-d. 

Felt free to get rid of this unnecessary check and make the code look cleaner.

Will revert this commit if this change is unneeded.

pkg/controller/backupbucket/actuator.go
renormalize: Sorry for this oversight!

Yes, of course. I've made a similar error before in one of my previous PRs and you'd suggested the same. Thanks for reminding.

pkg/controller/backupbucket/actuator.go
renormalize: I agree with the suggestion to move the `// lock the policy` section to `updateBackupBucketIfNeeded` as well. Will do so in the next commit.
I'd restructured the code to take this form before raising the PR since the GCP PR had formatted it this way.

Once the lock section is moved to `updateBackupBucketIfNeeded`, we could avoid the second GET - the call `CreateOrUpdateImmutabilityPolicy` will return an etag in its response, and this can be used for the `LockImmutabilityPolicy` call. Will make this change in the next commit. Thanks for the suggestion!

pkg/controller/backupbucket/actuator.go
renormalize: This part is reached logically in these cases:

- An immutability policy is being added to a bucket without such a policy. This policy can be `locked: false` or 'locked: true`.
- An immutability policy is being **updated** for a bucket which already contains an **unlocked** policy. This could mean a reduction of immutability duration or extension of immutability duration. (Not deletion, as that is done through a different API call).

If the policy is locked, then we would not be using this method. To extend the duration of a locked policy, we invoke `ExtendImmutabilityPolicy`, and obviously a locked policy can not be deleted.

pkg/controller/backupbucket/actuator.go
renormalize: https://github.com/gardener/gardener-extension-provider-azure/pull/1098#discussion_r1971931335

pkg/controller/worker/machines.go
kon-angelo: Let me doublecheck that then and reply shortly 👍

pkg/controller/backupbucket/actuator.go
kon-angelo: Thank you for the explanation!

pkg/controller/backupbucket/actuator.go
renormalize: Addressed in [ff6ad6e](https://github.com/gardener/gardener-extension-provider-azure/pull/1098/commits/ff6ad6ea46497187dff5f1accedf57a43c2a87bb).

pkg/controller/backupbucket/actuator.go
renormalize: This comment was removed since the additional GET call is now avoided.

pkg/admission/validator/seed.go
renormalize: Addressed in [bfa7104](https://github.com/gardener/gardener-extension-provider-azure/pull/1098/commits/bfa71047fe56a03414a08345481d9b18de23e22b).

pkg/controller/backupbucket/actuator.go
renormalize: Addressed in [bfa7104](https://github.com/gardener/gardener-extension-provider-azure/pull/1098/commits/bfa71047fe56a03414a08345481d9b18de23e22b).

renormalize:
It was noticed that control plane migrations of shoots scheduled on seeds with immutable buckets would fail since the `BackupEntry`s are deleted forcefully during migrations; see [this](https://github.com/gardener/gardener/blob/e8ec99ee8d323371052639385c7b6f7cc7fe4b8f/pkg/gardenlet/controller/shoot/shoot/reconciler_reconcile.go#L284-L289) and [this](https://github.com/gardener/gardener/blob/e8ec99ee8d323371052639385c7b6f7cc7fe4b8f/pkg/gardenlet/operation/botanist/backupentry.go#L85-L92). A fix was planned and the corresponding PR was raised: https://github.com/gardener/gardener/pull/11388.

However, after more deliberation and research, it was concluded that the proposed changes in the gardener/gardener PR are not the optimal way to solve this problem, as hinted at by the review comments in the gardener/gardener PR.

The workaround solution that I propose is to use [data lifecycle management policies](https://learn.microsoft.com/en-us/azure/storage/blobs/lifecycle-management-overview) provided by Azure. 
The gist of this solution is that we set a lifecycle policy on the bucket, which deletes objects which meet certain conditions.
The policy will act on objects which are tagged with certain tags, which are added to the objects during deletion of the `BackupEntry`.

This solution is a delayed-delete, which is performed by the infrastructure provider completely by itself.

With this approach, we satisfy the immutability properties of objects in the bucket, and avoid making changes to gardener/gardener to make it work with buckets with certain configurations.

****

I've yet to implement these changes in this PR. However, the changes will be quite simple. I estimate that I will be able to push them sometime tomorrow.

renormalize:
This solution was fully agreed upon by the etcd-druid maintainers yesterday, therefore I was not able to include these changes in the initial commits with which I raised this PR.

However, the changes will be quite simple. I estimate that I will be able to push them sometime tomorrow.

docs/usage/usage.md
```diff
@@ -701,3 +701,48 @@ Some key facts about VMSS Flex based clusters:
 - In case the configuration of the VMSS will change (e.g. amount of fault domains in a region change; configured in the CloudProfile) all machines of the worker pool need to be rolled
 - It is not possible to migrate an existing primary AvailabilitySet based Shoot cluster to VMSS Flex based Shoot cluster and vice versa
 - VMSS Flex based clusters are using `Standard` SKU LoadBalancers instead of `Basic` SKU LoadBalancers for AvailabilitySet based Shoot clusters
+
+## `BackupBucketConfig`
+
+`BackupBucketConfig` describes the configuration that needs to be passed over for creation of the backup bucket infrastructure. Configuration like immutability (WORM, i.e. write-once-read-many) that can be set on the bucket are specified here. Objects in the bucket will inherit the immutability duration which is set on the bucket, and they can not be modified or deleted for that duration.
+
+This extension supports creating (and migrating already existing buckets) to use [container-level WORM policies](https://learn.microsoft.com/en-us/azure/storage/blobs/immutable-container-level-worm-policies).
```
ishan16696, line 709:
```suggestion
This extension supports creating (and migrating already existing buckets if enabled) to use [container-level WORM policies](https://learn.microsoft.com/en-us/azure/storage/blobs/immutable-container-level-worm-policies).
```

docs/usage/usage.md
```diff
@@ -701,3 +701,48 @@ Some key facts about VMSS Flex based clusters:
...
+Example:
+
+```yaml
+apiVersion: azure.provider.extensions.gardener.cloud/v1alpha1
+kind: BackupBucketConfig
+immutability:
+  retentionType: "bucket"
+  retentionPeriod: 24h
+  locked: false
+```
```
ishan16696, line 720:
```suggestion
```yaml
kind: BackupBucketConfig
immutability:
  retentionType: "bucket"
  retentionPeriod: 24h
  lockedPolicy: false
```
```

Can we use `lockedPolicy` instead of just `locked` as to me `lockedPolicy` name makes more sense ?

cc @seshachalam-yv @kon-angelo

File: docs/usage/usage.md [hunk 11]
```diff
@@ -701,3 +701,48 @@ Some key facts about VMSS Flex based clusters:
...
+    name: my-azure-secret
+    namespace: my-namespace
+  providerConfig:
+    apiVersion: azure.provider.extensions.gardener.cloud/v1alpha1
+    kind: BackupBucketConfig
+    immutability:
+      retentionType: "bucket"
+      retentionPeriod: 24h
+      locked: true
+```
```
ishan16696, line 748: I checked the `BackupBucket ` resource, I didn't find `.spec.provider.region`. I find this `.spec.region`, can you please check this again , what's the correct resource ?
```yaml
apiVersion: extensions.gardener.cloud/v1alpha1
kind: BackupBucket
metadata:
...
spec:
  region: westeurope
  secretRef:
    name: ..
    namespace: garden
  type: azure
```

pkg/admission/validator/seed_test.go
```diff
@@ -0,0 +1,266 @@
...
+			Entry("Backup not configured",
+				generateSeed("", "", false, false),
+				generateSeed("", "", false, false),
+			),
+		)
+
+		DescribeTable("Invalid update scenarios",
+			func(oldSeed, newSeed *core.Seed, expectedError string) {
+				err := seedValidator.Validate(context.Background(), newSeed, oldSeed)
+				fmt.Printf("%#v\n", err)
```
ishan16696, line 141: may be you forgot to remove this ?

ishan16696 (changes requested):
few nits:

pkg/admission/validator/seed_test.go
renormalize: Will remove in a future commit where newer changes will be introduced. Thanks for noticing.

docs/usage/usage.md
kon-angelo: I don't have a strong opinion here. For a boolean value `locked` is okay. 
It will diverge from the GCP spec but let's decide on the correct name and it can be retrofitted if needed.

pkg/admission/validator/seed_test.go
renormalize: Addressed in [6c7a4c](https://github.com/gardener/gardener-extension-provider-azure/pull/1098/commits/36c7a4c9b34af623d81c5831b8bda99343c0f48a).

docs/usage/usage.md
renormalize: Addressed in [6c7a4c](https://github.com/gardener/gardener-extension-provider-azure/pull/1098/commits/36c7a4c9b34af623d81c5831b8bda99343c0f48a).

docs/usage/usage.md
renormalize: Addressed in [6c7a4c](https://github.com/gardener/gardener-extension-provider-azure/pull/1098/commits/36c7a4c9b34af623d81c5831b8bda99343c0f48a).

docs/usage/usage.md
```diff
@@ -701,3 +701,48 @@ Some key facts about VMSS Flex based clusters:
 - In case the configuration of the VMSS will change (e.g. amount of fault domains in a region change; configured in the CloudProfile) all machines of the worker pool need to be rolled
 - It is not possible to migrate an existing primary AvailabilitySet based Shoot cluster to VMSS Flex based Shoot cluster and vice versa
 - VMSS Flex based clusters are using `Standard` SKU LoadBalancers instead of `Basic` SKU LoadBalancers for AvailabilitySet based Shoot clusters
+
+## `BackupBucketConfig`
```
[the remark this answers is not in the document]
renormalize: Addressed in [6c7a4c](https://github.com/gardener/gardener-extension-provider-azure/pull/1098/commits/36c7a4c9b34af623d81c5831b8bda99343c0f48a).

pkg/controller/backupbucket/actuator.go
```diff
@@ -58,34 +59,138 @@ func (a *actuator) Reconcile(ctx context.Context, _ logr.Logger, backupBucket *e
...
 
-	blobStorageClient, err := DefaultBlobStorageClient(ctx, a.client, backupBucket.Status.GeneratedSecretRef)
+	// add lifecycle policies to the storage account to perform delayed delete of backupentries
+	managementPoliciesClient, err := factory.ManagementPolicies()
 	if err != nil {
 		return util.DetermineError(err, helper.KnownCodes)
 	}
-	return util.DetermineError(blobStorageClient.CreateContainerIfNotExists(ctx, backupBucket.Name), helper.KnownCodes)
+	if err = managementPoliciesClient.CreateOrUpdate(ctx, resourceGroupName, storageAccountName, 0); err != nil {
+		logger.Error(err, "Failed to add the lifecycle policy on the storage account")
```
seshachalam-yv, line 99:
```suggestion
		logger.Error(err, "Failed to add or update the lifecycle policy on the storage account")
```

seshachalam-yv (commented):
Overall Looks good to me.

pkg/azure/client/storage.go
```diff
@@ -86,55 +87,58 @@ func NewBlobStorageClientFromSecretRef(ctx context.Context, client client.Client
...
-func (c *BlobStorageClient) DeleteContainerIfExists(ctx context.Context, container string) error {
-	_, err := c.client.DeleteContainer(ctx, container, nil)
-	if err == nil || bloberror.HasCode(err, bloberror.ContainerBeingDeleted, bloberror.ContainerNotFound) {
-		return nil
+	// Blob is immutable, set the tag that triggers cleanup through lifecycle policy
+	if bloberror.HasCode(err, bloberror.BlobImmutableDueToPolicy) {
+		tags := map[string]string{
+			azure.BlobMarkedForDeletionTagKey: "true",
+		}
+		_, err = blobClient.SetTags(ctx, tags, nil)
```
seshachalam-yv, line 141: Don't set if already blob contains tag `azure.BlobMarkedForDeletionTagKey`

pkg/controller/backupbucket/actuator.go
```diff
@@ -58,34 +59,138 @@ func (a *actuator) Reconcile(ctx context.Context, _ logr.Logger, backupBucket *e
...
 	}
 
-	blobStorageClient, err := DefaultBlobStorageClient(ctx, a.client, backupBucket.Status.GeneratedSecretRef)
+	// add lifecycle policies to the storage account to perform delayed delete of backupentries
+	managementPoliciesClient, err := factory.ManagementPolicies()
 	if err != nil {
 		return util.DetermineError(err, helper.KnownCodes)
 	}
-	return util.DetermineError(blobStorageClient.CreateContainerIfNotExists(ctx, backupBucket.Name), helper.KnownCodes)
+	if err = managementPoliciesClient.CreateOrUpdate(ctx, resourceGroupName, storageAccountName, 0); err != nil {
```
kon-angelo, line 98: - Do we always now (and unconditionally) create management policies for our buckets ? 
- Does that 0 indicate immediate deletion ? As in, immediately delete anything that has that tag?

pkg/controller/backupbucket/actuator.go
renormalize: Will add it in the next commit!

pkg/controller/backupbucket/actuator.go
renormalize: Thanks for the review @kon-angelo!

> Do we always now (and unconditionally) create management policies for our buckets ?

Yes. We add this unconditionally to the storage accounts.

Why we would not want to do this:

- If the bucket is not immutable, the management policy would never take affect, since we directly delete the objects in the bucket when the `BackupEntry` is deleted. In this case, the policy is not needed at all.

Why we would want to do this.

- The lifecycle policy could first be checked if it existed, and then if it is not present, the policy can be added. The type of lifecycle policy that we want on the storage account is *always the same*. It is simpler to just establish that by writing the policy on every reconcile.
- If the bucket is made immutable, the only possible way to delete the `BackupEntry` while respecting the immutability of the objects is to add blob tags to the objects, and make the lifecycle policy perform the delayed delete on the objects with these tags (which correspond to the now deleted `BackupEntry`).
- Consider a bucket that was originally immutable, and the immutability is now switched off. If the lifecycle policy is also removed along with immutability, it is almost a certainty that objects that were going to be deleted by the lifecycle policy will now never be deleted.
These objects will forever consume space in the bucket, wasting resources; and also will require manual intervention from operators to clean them up.

With the above context, it would be better to **always add the policy** to the storage account, whether the bucket is immutable or not to ensure resources are not wasted.

I wouldn't mind first fetching the policy on the storage account, and then adding it if it does not exist. I will make that change if you think thats' the better way of doing it.

****

> Does that 0 indicate immediate deletion ? As in, immediately delete anything that has that tag?

The 0 here indicates that the lifecycle policy will attempt to delete any object with the tags 0 days after the object's creation. The policy's attempt to delete objects will fail on the provider's side until the objects' immutability expires.

Once the immutability expires, the policy to delete the objects will succeed, and the objects are finally removed.

Instead of passing `0`, the number of days the bucket is immutable for can also be passed here as the effect would be the same, but the `BackupEntry` does not have access to the `providerConfig` of the `BackupBucket`. I do not want to couple `BackupEntry` and `BackupBucket` resources just for this configuration.

****

> Does this require additional permissions from the ones we currently use ?

Yes.
Consumers of this extension would need to configure their Azure permissions to allow these API endpoints:

```
Microsoft.Storage/storageAccounts/managementPolicies/read
Microsoft.Storage/storageAccounts/managementPolicies/write
Microsoft.Storage/storageAccounts/managementPolicies/delete
```

The code currently only will use `/write`, but it would be easy to configure all three at once.

We would also need to add 

```
Microsoft.Storage/storageAccounts/blobServices/containers/immutabilityPolicies/*
```

to the permissions to make any of the immutability features work.

kon-angelo: Although the distinction between seed and shoot permissions is not strictly maintained in this https://github.com/gardener/gardener-extension-provider-azure/blob/master/docs/usage/azure-permissions.md#microsoftstorage , you should still add these permissions there. 

In addition, you need a breaking change notice for this PR. Any operator updating the extension effectively needs to change his permissions regardless of whether he is using immutable buckets or not.

renormalize: Yes, it was a miss from my side to update the documentation to include these permissions. Thanks for pointing it out.
I will update the docs, and modify the release notes of the PR to notify of the breaking change.

pkg/controller/backupbucket/actuator.go
renormalize: Addressed in [c05eb7f](https://github.com/gardener/gardener-extension-provider-azure/pull/1098/commits/c05eb7f9bf5ea9ba38863ba5648c4360329d0f40).

pkg/controller/backupbucket/actuator.go
renormalize: Updated the documentation in [f81ed04](https://github.com/gardener/gardener-extension-provider-azure/pull/1098/commits/f81ed0402230e9e708152b52aa579ddeaae131b1).

kon-angelo (dismissed):
/lgtm

you need to run the make generate though

kon-angelo (approved):
/lgtm

PR #1114, opened 2025-03-13
Implement etcd credential rotation

kon-angelo:

/area control-plane
/kind enhancement
/platform azure

Fixes #

**Release note**:

```other operator
Implement etcd credential rotation
```


renormalize (changes requested):

pkg/azure/client/storageaccount.go
```diff
@@ -29,18 +29,27 @@ func NewStorageAccountClient(auth *internal.ClientAuth, tc azcore.TokenCredentia
...
+// CreateOrUpdateStorageAccount creates a storage account.
+func (c *StorageAccountClient) CreateOrUpdateStorageAccount(ctx context.Context, resourceGroupName, storageAccountName, region string, keyExpiration *int32) error {
+	properties := armstorage.AccountPropertiesCreateParameters{
+		AccessTier:             ptr.To(armstorage.AccessTierCool),
+		EnableHTTPSTrafficOnly: ptr.To(true),
+		AllowBlobPublicAccess:  ptr.To(false),
+		MinimumTLSVersion:      ptr.To(armstorage.MinimumTLSVersionTLS12),
+		KeyPolicy: &armstorage.KeyPolicy{
+			KeyExpirationPeriodInDays: ptr.To(int32(0)),
 		},
```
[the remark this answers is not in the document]
kon-angelo: Well this was my first thought too. Upon experimentation (due to the lack of documented behavior) I came to the following conclusions:
- Setting it to 0, basically unsets the expiration period.
- Without explicitly setting it to 0 you cannot "unset" the expiration. This is not typical of the usual azure SDK behavior that usually relies on "update", rather than "patch" calls. In this case leaving this to nil does not make any changes to the bucket.

pkg/controller/backupbucket/backupbucket.go
```diff
@@ -23,33 +30,113 @@ func GenerateStorageAccountName(backupBucketName string) string {
...
+func ensureKeyRotated(
+	ctx context.Context,
+	log logr.Logger,
+	factory azureclient.Factory,
+	resourceGroupName, storageAccountName, currentKey string,
+	bb *extensionsv1alpha1.BackupBucket,
+	bbCfg *azure.BackupBucketConfig,
+) (*armstorage.AccountKey, bool, error) {
+	// skip rotation if not configured
+	if bbCfg.RotationConfig == nil || bbCfg.RotationConfig.RotationPeriodDays == 0 {
```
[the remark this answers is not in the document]
kon-angelo: Yes, we should never really reach this point. I will to have a couple extra defences in my coding. Maybe returning an error instead is better as that would notify us that something outside our expectations happened ?

pkg/azure/client/storageaccount.go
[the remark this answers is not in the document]
renormalize: Got it. Thanks for the explanation!

pkg/controller/backupbucket/backupbucket.go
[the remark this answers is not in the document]
renormalize: Understood. An error returned here is the better idea.

pkg/controller/backupbucket/actuator.go
```diff
@@ -132,14 +118,78 @@ func (a *actuator) Reconcile(ctx context.Context, logger logr.Logger, backupBuck
...
+					delete(backupBucket.GetAnnotations(), azuretypes.StorageAccountKeyMustRotate)
+					if err := a.client.Patch(ctx, backupBucket, backupBucketPatch); err != nil {
+						return logWithError(logger, err, "Failed to remove the rotation annotation")
+					}
+				}
+
+				logger.Info("Updating backupbucket with new account key", "name", *storageAccountKey.KeyName)
+				if err := a.createOrUpdateBackupBucketGeneratedSecret(ctx, backupBucket, storageAccountName, *storageAccountKey.Value, storageDomain); err != nil {
+					return logWithError(logger, err, "Failed to generate the backupbucket secret")
+				}
```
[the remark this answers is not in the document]
renormalize: It was conveyed by @/kon-angelo that removal of the annotation before update of the secret would be the better approach, if the annotation is to be supported.

If the removal of the annotation fails after credential rotation, the credentials will be unintentionally rotated in the next reconciliation, which might cause things to break.

pkg/controller/backupbucket/backupbucket.go
```diff
@@ -23,33 +30,124 @@ func GenerateStorageAccountName(backupBucketName string) string {
...
+			return nil, err
+		}
+		if v, ok := secret.Data[azuretypes.StorageKey]; ok {
+			// The key in the secret can either be the most recent, the oldest, or not found in the list of keys from Azure API.
+			// Unless the secret value matches the most recent key, we can assume that the rotation already happened but the secret was not properly
+			// updated. Hence, we no-op and return the current most recent key.
+			if string(v) != *mostRecentKey.Value {
+				log.Info("Skipping rotation because backup bucket current key does not match storage account keys. Using most recent rotation key instead.")
+				return currentKeys, nil
+			}
```
renormalize, line 134: Nice fix to the potential double rotation for a single annotation!

pkg/azure/client/storageaccount.go
```diff
@@ -52,21 +60,30 @@ func (c *StorageAccountClient) CreateStorageAccount(ctx context.Context, resourc
...
+		nil,
+	)
+
+	if err != nil {
+		return nil, err
 	}
 
-	firstKey := response.Keys[0]
-	return *firstKey.Value, nil
+	return resp.Keys, nil
```
renormalize, line 88: Nit: If the client methods always return the sorted key slice, there is no need to explicitly sort the keys elsewhere in the code, making it easier to read.
However, if you feel explicitly sorting the keys specifies intent, then we can keep the code as is.

```suggestion
	return SortKeysByAge(resp.Keys), nil
```

pkg/azure/client/storageaccount.go
```diff
@@ -52,21 +60,30 @@ func (c *StorageAccountClient) CreateStorageAccount(ctx context.Context, resourc
...
+	response, err := c.client.ListKeys(ctx, resourceGroupName, storageAccountName, &armstorage.AccountsClientListKeysOptions{})
 
 	if err != nil {
-		return "", err
+		return nil, err
 	}
 
-	if len(response.Keys) < 1 {
-		return "", fmt.Errorf("no key found in storage account %s", storageAccountName)
+	return response.Keys, nil
```
renormalize, line 71: https://github.com/gardener/gardener-extension-provider-azure/pull/1114#discussion_r2011774998

```suggestion
	return SortKeysByAge(response.Keys), nil
```

pkg/controller/backupbucket/actuator.go
```diff
@@ -132,14 +116,43 @@ func (a *actuator) Reconcile(ctx context.Context, logger logr.Logger, backupBuck
...
+	}
+	if len(keys) < 1 {
+		return logWithError(log, nil, "Storage account did not return any keys")
+	}
+
+	if keys, err = a.ensureKeyRotated(ctx, log, storageAccountClient, resourceGroupName, storageAccountName, keys, backupBucket, backupBucketConfig); err != nil {
+		return logWithError(log, err, "Failed to ensure account key rotation")
+	}
+
+	if err := a.createOrUpdateBackupBucketGeneratedSecret(ctx, backupBucket, storageAccountName, *SortKeysByAge(keys)[0].Value, storageDomain); err != nil {
```
renormalize, line 150: https://github.com/gardener/gardener-extension-provider-azure/pull/1114#discussion_r2011775725

pkg/controller/backupbucket/backupbucket.go
```diff
@@ -23,33 +30,124 @@ func GenerateStorageAccountName(backupBucketName string) string {
...
+	if backupBucketConfig.RotationConfig == nil {
+		log.Info("Skipping rotation because it's not configured for the backup bucket")
+		return currentKeys, nil
+	}
+	// we should never enter this condition as it is protected by the admission controller. Still we assert our invariants.
+	if backupBucketConfig.RotationConfig.RotationPeriodDays == 0 {
+		log.Error(nil, "backup bucket rotation config is required")
+		return currentKeys, nil
+	}
+	mostRecentKey := SortKeysByAge(currentKeys)[0]
```
renormalize, line 109: https://github.com/gardener/gardener-extension-provider-azure/pull/1114#discussion_r2011775725

renormalize (approved):
Thanks for making the changes in the latest commits @kon-angelo!

AndreasBurger (approved):
lgtm

pkg/apis/azure/validation/backupbucket.go
```diff
@@ -9,31 +9,63 @@ import (
...
 		return allErrs
 	}
+	if cfg.RotationPeriodDays < 1 {
+		return append(allErrs, field.Required(fldPath.Child("rotationPeriodDays"), "rotationPeriod must be configured if key rotation is enabled"))
+	}
 
-	if backupBucketConfig.Immutability.RetentionType != apisazure.BucketLevelImmutability {
-		allErrs = append(allErrs, field.Invalid(fldPath.Child("immutability", "retentionType"), backupBucketConfig.Immutability.RetentionType, "must be 'bucket'"))
+	if cfg.RotationPeriodDays < 2 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("retentionPeriodDays"), cfg.RotationPeriodDays, "must be greater than 2 days"))
```
hebelsan, line 39:
```suggestion
		allErrs = append(allErrs, field.Invalid(fldPath.Child("rotationPeriodDays"), cfg.RotationPeriodDays, "must be greater than 2 days"))
```

pkg/apis/azure/validation/backupbucket.go
```diff
@@ -9,31 +9,63 @@ import (
...
 
-	if backupBucketConfig == nil || backupBucketConfig.Immutability == nil {
+	return allErrs
+}
+func validateKeyRotation(cfg *apisazure.RotationConfig, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if cfg == nil {
 		return allErrs
 	}
+	if cfg.RotationPeriodDays < 1 {
```
hebelsan, line 34: A Nit but do we really need this check?
Because one line below we check for `if cfg.RotationPeriodDays < 2`

PR #1122, opened 2025-03-21
Update worker pool hash if diagnosticProfile is enabled

hebelsan:

/area usability
/kind enhancement
/platform azure

Update worker pool hash if diagnosticProfile is enabled to roll nodes.
Before this PR nodes would only be rolled if diagnosticProfile is enabled and a storageURI is specified.
However the more common approach is to leave the storageURI empty and let Azure handle the storage.

Fixes #

**Release note**:

```other user
Update worker pool hash if diagnostic profile option is enabled
```

### DIFF
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -538,9 +538,9 @@ func (w workerDelegate) workerPoolHashDataV2(workerConfig api.WorkerConfig, addi
 
 	if workerConfig.DiagnosticsProfile != nil && workerConfig.DiagnosticsProfile.Enabled {
 		hashData = append(hashData, "DiagnosticsProfileEnabled")
-	}
-	if workerConfig.DiagnosticsProfile != nil && workerConfig.DiagnosticsProfile.StorageURI != nil {
-		hashData = append(hashData, *workerConfig.DiagnosticsProfile.StorageURI)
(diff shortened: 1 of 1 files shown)


pkg/controller/worker/machines.go
```diff
@@ -534,9 +534,11 @@ func (w *workerDelegate) generateWorkerPoolHash(pool extensionsv1alpha1.WorkerPo
...
 // workerPoolHashDataV2 adds additional provider-specific data points to consider to the given data.
 func (w workerDelegate) workerPoolHashDataV2(workerConfig api.WorkerConfig, additionalData []string) ([]string, error) {
-	hashData := make([]string, len(additionalData))
-	copy(hashData, additionalData)
+	hashData := append([]string{}, additionalData...)
 
+	if workerConfig.DiagnosticsProfile != nil && workerConfig.DiagnosticsProfile.Enabled {
+		hashData = append(hashData, "DiagnosticsProfileEnabled")
+	}
 	if workerConfig.DiagnosticsProfile != nil && workerConfig.DiagnosticsProfile.StorageURI != nil {
```
kon-angelo, line 542: I think this can be included in the condition above

kon-angelo (approved):
/lgtm

PR #1143, opened 2025-04-14
Correctly mount workload identity token to csi driver controller

dimityrmirchev:

/area security
/kind bug
/platform azure

Fixes #

**Release note**:

```bugfix user
An issue causing `csi-driver-controller` to not have mounted a workload identity token when the feature is enabled is now fixed.
```


ialidzhikov (approved):
/lgtm

PR #1211, opened 2025-07-07
Set user agent when creating Azure clients

AndreasBurger:

/area audit-logging
/kind enhancement
/platform azure

Adds information to the UserAgent string of the created client to make it easier to identify calls made by clients created by the extension provider

Fixes #

Azure seems to have the most restricting of the Provider-SDKs. I've decided to leave the other UA-strings consistent and just shorten the Azure one, but I'll need to test how strictly this is actually enforced.

**Release note**:

```other operator
Clients created by the Azure extension provider will now identify themselves by adding to the `user-agent` header of their calls.
```


kon-angelo (approved):
/lgtm 

can we progress with the PR

hebelsan (approved):
/lgtm

PR #1245, opened 2025-08-12
Fix race condition in BackupBucket integration test

wpross:

/area testing
/kind flake
/platform azure

Fixes possible race condition for the key rotation test:
Instead of waiting for a successful reconciliation of the backupbucket, wait for 
1. observed generation to match generation and 
2. for the rotation annotation to be removed 
before checking the rotated keys. The 1st check ensures that the backupbucket is reconciled with the patched version that includes the rotation annotation, and the 2nd check ensures that the rotation actually happened. 

Fixes possible issues when removing the immutability policy:
Use path instead of update to make sure the policy is removed from the spec.
Check based on observed generation if the changes have been applied, then wait for the backupbucket reconciliation to finish and then wait until the policy is actally removed on the azure container. 

Fixes #

**Release note**:

```other operator

```


hebelsan (approved):
/lgtm

PR #1265, opened 2025-08-29
[GEP-26] Add support for Workload Identity credentials to etcd backups

vpnachev:
/area security ipcei documentation
/kind enhancement
/platform azure

[GEP-26] Add support for Workload Identity credentials to etcd backups

Part of https://github.com/gardener/gardener/issues/9586

cc @dimityrmirchev 

**Release note**:

```feature operator
This extension now supports `WorkloadIdentity`s as credentials for etcd backup.
```


pkg/apis/azure/helper/scheme.go
```diff
@@ -163,3 +163,15 @@ func InfrastructureStatusFromInfrastructure(infra *extensionsv1alpha1.Infrastruc
...
 	return status, nil
 }
+
+// WorkloadIdentityConfigFromBytes extracts WorkloadIdentityConfig from the provided byte array.
+func WorkloadIdentityConfigFromBytes(config []byte) (*api.WorkloadIdentityConfig, error) {
+	if len(config) == 0 {
+		return nil, fmt.Errorf("cannot parse WorkloadIdentityConfig from empty config")
+	}
+	workloadIdentityConfig := &api.WorkloadIdentityConfig{}
+	if _, _, err := decoder.Decode(config, nil, workloadIdentityConfig); err != nil {
```
dimityrmirchev, line 173: Should we use util.Decode here?

```
"github.com/gardener/gardener/extensions/pkg/util"
```

dimityrmirchev (commented):
One question, other looks good

pkg/apis/azure/helper/scheme.go
dimityrmirchev: > We already agreed on this approach in provider-aws

That change I did not see 🙈 

> Is there any reason to use it? 

I guess the benefit is this comment https://github.com/gardener/gardener/blob/56037371cdd961ee894592c6bcaf101d0aabe9bc/extensions/pkg/util/serialization.go#L17-L19

TBH I see that both are being used through the repository and I am not sure when to use which. My guess is that util.Decode is used during validation to ensure proper api versioning. The contra question would be "Why introduce the change in the first place? What was the benefit?"

vpnachev:
Just force pushed to resolve conflicts, no actual change in the PR.

Makefile
```diff
@@ -24,6 +24,7 @@ WEBHOOK_CONFIG_PORT	:= 8443
 WEBHOOK_CONFIG_MODE	:= url
 WEBHOOK_CONFIG_URL	:= host.docker.internal:$(WEBHOOK_CONFIG_PORT)
 EXTENSION_NAMESPACE	:=
```
hebelsan, line 26: Actually, why is the `EXTENSION_NAMESPACE` set twice?

pkg/azure/client/auth.go
```diff
@@ -62,10 +63,34 @@ func GetClientAuthData(ctx context.Context, c client.Client, secretRef corev1.Se
...
+			if !ok {
+				return nil, fmt.Errorf("secret %q is missing a 'config' data key", client.ObjectKeyFromObject(secret).String())
+			}
+
+			workloadIdentityConfig, err := helper.WorkloadIdentityConfigFromBytes(config)
+			if err != nil {
+				return nil, fmt.Errorf("could not decode 'config' as WorkloadIdentityConfig: %w", err)
+			}
+
+			subscriptionID = workloadIdentityConfig.SubscriptionID
```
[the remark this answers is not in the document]
vpnachev: The admission controller ensures that the fields are set, see https://github.com/gardener/gardener-extension-provider-azure/blob/3be4243bb37321d92339f9308c72889bdd4686b5/pkg/apis/azure/validation/workloadidentity.go#L18-L26

hebelsan (commented):
/lgtm

PR #1271, opened 2025-09-04
Add stable outbound rules for loadbalancer

kon-angelo:

/area control-plane
/kind enhancement
/platform azure

Allows to configure a shoot cluster with a loadbalancer and outbound rules. This has the follwing benefits:
- Allows shoot creation with LB-based outbound traffic (meaning it can work post 30/09/25 Azure change)
- Stable IPs for Egress. Meaning that we can expose `EgressCIDRs` even with LB-based outbound traffic.

Transition to the current LB setup and NATGW works in both forward, but backwards requires a node rollout.

image

Depends on https://github.com/gardener/machine-controller-manager-provider-azure/pull/199

Fixes #

**Release note**:

```other operator

```


adenitiu:
@kon-angelo  was this intentionally left to close, or will there be a different approach taken ?

kon-angelo:
@adenitiu it may indeed make a comeback in the future but https://github.com/gardener/gardener-extension-provider-azure/pull/1469 should be the easiest to implement for our use case

PR #1354, opened 2025-11-05
Fix PR milestone check workflow

wpross:
* make sure check is always based on latest PR data
* add check to ensure correct milestone is set
* get version from master branch

/area dev-productivity
/kind bug
/platform azure

What this PR does / why we need it:
this PR ensures that when a milestone of a PR is checked, that the check is always based on the latest data. So when the check fails since e.g. no milestone is set on a PR and then the milestone is manually set and the check re-run, it should now succeed.

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/.github/workflows/release-milestone.yaml
+++ b/.github/workflows/release-milestone.yaml
@@ -7,22 +7,25 @@ on:
     - synchronize
     - labeled
     - unlabeled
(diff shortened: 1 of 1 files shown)


hebelsan (approved):
/lgtm

PR #1382, opened 2025-11-23
force read cache for in-tree pvs annotation

kon-angelo:

/area control-plane
/kind enhancement
/platform azure

Adds an annotation to force attach in-tree PVs in readOnly mode. Before that they were attached in ReadWrite which could lead to issue for certain applications. While CSI PV spec does allow you to pick the value, this is not support for unmigrated in-tree PVs. The annotation added by the upstream project in https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/3414 allows for a way to override the RW behavior irrespective of the PV spec.

Fixes #

cc: @dguendisch 

**Release note**:

```other operator
Annotations added to ensure in-tree PVs can be forced attached to the node in ReadOnly caching mode.
```

### DIFF
--- a/pkg/controller/controlplane/valuesprovider.go
+++ b/pkg/controller/controlplane/valuesprovider.go
@@ -598,7 +598,7 @@ func getCSIControllerChartValues(
 	infraStatus *apisazure.InfrastructureStatus,
 	checksums map[string]string,
 	useWorkloadIdentity bool,
-) (map[string]interface{}, error) {
+) (map[string]any, error) {
 	values := map[string]any{
 		"enabled": true,
 		"podAnnotations": map[string]any{


pkg/controller/controlplane/valuesprovider_test.go
```diff
@@ -435,6 +435,45 @@ var _ = Describe("ValuesProvider", func() {
...
+			shootAnnotations := map[string]string{
+				azure.ShootDiskConvertRWCachingModeAnnotation: "true",
+			}
+			cluster = generateCluster(cidr, k8sVersion, false, shootAnnotations, nil, &gardencorev1beta1.Seed{})
+
+			cp := generateControlPlane(controlPlaneConfig, infrastructureStatus)
+			values, err := vp.GetControlPlaneChartValues(ctx, cp, cluster, fakeSecretsManager, checksums, false)
+
+			Expect(err).NotTo(HaveOccurred())
+			Expect(values).To(Equal(map[string]interface{}{
```
wpross: Maybe focus only on the values of interest so that the test won't break in case of changes on other fields.

wpross: probably better to use `any` instead of `interface{}` here and in the following.

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -609,6 +610,12 @@ func getCSIControllerChartValues(
 		},
 		"useWorkloadIdentity": useWorkloadIdentity,
 	}
+	convertRWCachingModeForInTreePV, _ := strconv.ParseBool(cluster.Shoot.Annotations[azure.ShootDiskConvertRWCachingModeAnnotation])
+	if convertRWCachingModeForInTreePV {
+		values["diskConfig"] = map[string]interface{}{
```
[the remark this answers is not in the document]
kon-angelo: quite correct. Missed that 😅

PR #1403, opened 2025-12-18
Fix: Support 2 key variants for dns secret keys

wpross:

/area robustness
/kind bug
/platform azure

This fix is to support not only the DNS-specific keys for dns provider secrets, but in addition the infrastructure secret keys.

Fixes #

**Release note**:

```bugfix operator
Support not only the DNS-specific keys for dns provider secrets, but in addition the infrastructure secret keys.
```


kon-angelo (approved):
/lgtm

PR #1411, opened 2026-01-02
Expose logging configuration in provider deployment

AndreasBurger:

/area logging
/kind task
/platform azure

Makes logging-related configuration accessible via chart-values.

Fixes #

Currently we only log at level 0 (info) and due to the way loggers are set up, "error" and "info" are the same level to the loggers. So for now this change could only be used to enable enable debug-level logging, but we do not yet emit debug-level logs.

Also fixes the rendering of the etcd-backup config in the provider-configmap

**Release note**:

```other operator
Extension-provider logging config can now be changed via helm-values
```
```breaking operator
ETCD backup config in the provider-chart no longer requires an additional `backup` yaml-level to render as expected
```


kon-angelo (approved):
/lgtm

PR #1549, opened 2026-05-29
fix: add missing get nodes permission to csi-driver-controller-disk

matthias-horne:

/area control-plane
/kind bug
/platform azure

This PR adds missing permissions to get `Nodes` to `csi-driver-controller-disk`. Without the permission the csi-driver-controller-disk fails with 
```
  E azure_controller_common.go:239 failed to get node info from labels:
  get node(...) failed with nodes "..." is forbidden:
  User "system:serviceaccount:kube-system:csi-driver-controller-disk"
  cannot get resource "nodes" in API group "" at the cluster scope
```
Error originates here: 
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/9a3763e7d9d546c98aa81f011ccf25a762bd7224/pkg/azuredisk/azure_controller_common.go#L236-L255

Adding the permission slightly changes the behavior in case more disks are attempted to be attached in a batch than luns are available on the node.

Before patch:
- node cannot be retrieved
- instance type is unknown
- max luns are unknown
- cannot get luns for all disks
- entire batch fails

After patch:
- retrieve instance type from  node labels
- get max luns for instance type
- get num attached disks on node
- calculate available luns
- reduce batch to num available luns
- attach the remaining disks

The network calls to the kube-apiserver or the provider API should not really change. The get nodes call to kube-apiserver will now succeed instead of fail. The `GetNodeDataDisks` to the provider API is now done earlier but since it is cached, this should not be an additional call.

Fixes #

**Release note**:

```bugfix user
Fix a RBAC permission issues that prevented batched disk attachments from partially succeeding if not enough free luns are available
```

## Summary by CodeRabbit

* **Bug Fixes**
  * CSI driver node controller now has appropriate permissions to access node information for improved disk management operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gardener/gardener-extension-provider-azure/pull/1549?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/charts/internal/shoot-system-components/charts/csi-driver-node/templates/clusterrole-csi-driver-controller-disk.yaml
+++ b/charts/internal/shoot-system-components/charts/csi-driver-node/templates/clusterrole-csi-driver-controller-disk.yaml
@@ -16,3 +16,6 @@ rules:
 - apiGroups: [""]
   resources: ["events"]
   verbs: ["create", "patch", "update"]
+- apiGroups: [""]
+  resources: ["nodes"]
+  verbs: ["get"]


hebelsan (commented):
/lgtm
/approve

